PR #12, opened 2022-03-09
#5: Added design document

### LINKED ISSUE #5: Create a design document for the Framework
# Background

- The framework should help with certain task and we need to document how the framework can help with them and what our assumptions are

# Acceptance Criteria

- Add a design document for the Framework
- Use OpenFastTrace https://github.com/itsallcode/openfasttrace

umitbuyuksahin:
fixes #5


doc/design.md
```diff
@@ -0,0 +1,69 @@
...
+available in Lua Scrips, `pquery` allows SQL queries to be run in the same 
+transactions.
+
+- The results of `pquery` are simple handles and there is a limit of 250 open 
+result set handles for Lua scripts.
+
+
+## Assumptions
+
+- Since the number open result set handles in Lua scripts is limited, it is
```
tkilias, line 22: this is also a constrain

doc/design.md
```diff
@@ -0,0 +1,69 @@
...
+- The results of `pquery` are simple handles and there is a limit of 250 open 
+result set handles for Lua scripts.
+
+
+## Assumptions
+
+- Since the number open result set handles in Lua scripts is limited, it is 
+assumed that Lua scripts is not nested more than 250 times.
+
+- The result of pquery is assumed to not exceed the limit in size.
```
tkilias, line 25:
```suggestion
- The result of pquery is assumed to not exceed the db memory limit in size. The DB usually allows 4 GB per query and node. The UDF can consume up to 32 GB per node, but get throttle down after the 4 GB. However, the 32 GB are shared between all queries on a node.
```

doc/design.md
```diff
@@ -0,0 +1,69 @@
...
+
```
tkilias, line 26: * The state of the Event Handler is not too large.
* The (de-)serialization and the upload/download of the state to the BucketFS take less than 3 seconds.

doc/design.md
```diff
@@ -0,0 +1,69 @@
...
+ 
+TBD 
+
+
+
+## Event Loop
+`dsn~event-loop~1`
+Event Loop  processes only the state transitions by executing queries returned 
+by the Event Handler. This loop provides the new state transition by calling 
+the Event Handler until the state is complete. Furthermore, the Event Loop
```
tkilias, line 49:
```suggestion
the Event Handler until the Event Handler returns the stop signal with a result or an error. Furthermore, the Event Loop 
```

doc/design.md
```diff
@@ -0,0 +1,69 @@
...
+* `feat~TBD~1`
+
+Needs: impl, utest, itest
+
+
+## Event Handler
+The Event Handler is a framework whose logic is implemented by users. It reacts 
+to the query results of  the Event Loop and generates SQL queries of the next 
+state. It stores the states in BucketFS.
+
```
tkilias, line 64: Yes, the event handler has two parts, the logic part is implemented by the user, the second part which calls the user code is implemented by us and provides certain features to the user code,  such as automatic state handling ((de-)serialization, upload, read). Managing temporary tables, which means the user code can register tables as temporary tables and the framework will remove them. It also forwards wraps the UDF context as iterable and forwards it to the user code. Furthermore, it builds the query to call the Event Handler again, from a given query which gives the result with which the Event handler will get called again.

tkilias (changes requested):
I would suggest, to maybe write the requirements first, this helps also with the design.

doc/design.md
```diff
@@ -0,0 +1,194 @@
...
+* `req~intiating-global-loop~1`
+
+Needs: impl, utest, itest
+
+### Handling States
+`dsn~handling-states~1`
+
+The  output of the called Event Handler query can include SQL queries to 
+complete its computation. In that case Event Loop run them on each iteration, 
+until the Event Handler returns the status as complete.
```
tkilias, line 73: The headline and the description don't fit, and later there is another section with storing the state in bucketfs. We only need to handle and store the state in the framework, because the execution of the event handler happens in a UDF and the UDF ends after each call to the event handler.

tkilias (changes requested):

doc/design.md
```diff
@@ -0,0 +1,194 @@
+# Design
+
+
+## Introduction
+This library is an event-driven framework that allows users to build complex
```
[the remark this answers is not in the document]
tkilias: We should introduce the framework more generally and then maybe state that we choose a event driven design because of the requirements and the limitation that pquery only is available for Lua at the moment.

PS: A dataflow based programming model/interface for the user is still possible. I will do something similar for machine learning. Where I build an execution graph and have an event driven graph runner which visits in the event handler the next node in the graph und returns its queries. I am only lazy regarding loops, which handle by calling another instance of this framework. However, this should be possible to incorporate as well. We might add this dataflow based programming model to this library, but not now.

doc/design.md
```diff
@@ -0,0 +1,194 @@
...
+`dsn~cleanup-temporary-bucketfs-files~1`
+
+Temporary BucketFS files are removed in case of an error that occur during 
+the execution of the framework with the keeping option  not selected.
+
+Covers:
+
+* `req~cleanup-temporary-bucketfs-files~1`
+
+Needs: impl, utest, itest
```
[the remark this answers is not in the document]
umitbuyuksahin: preparing a new language container could be challenge for some users. I'm in favor of creating python udf script.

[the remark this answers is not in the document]
tkilias: we need both here, for the machine learning library we will build a separate language container

doc/design.md
```diff
@@ -0,0 +1,249 @@
...
+Covers:
+
+* `req~managing-temporary-tables~1`
+
+Needs: impl, utest, itest
+
+
+### Electing Leader
+`dsn~electing-leader~1`
+
```
tkilias, line 193: Before leader election can happen. UDFs need to discover each other. We will use pycos, but it works, in general, the following. Every UDF receives on a fixed port UDP packages and sends UDP broadcasts on the same port. This way, everyone should find everyone. Assuming the Network has enough capacity left to send the UDP packages. To connect not to UDF instances of a different UDF call, each UDF call includes a unique ID which will be part of the broadcast message. We use this ID to connect only to other UDF instances of the same UDF Call.

Further, we know how many UDF instances should have been started, and we wait with leader election, until all instances have connected to each other.

tkilias (changes requested):

doc/design.md
```diff
@@ -0,0 +1,258 @@
...
+The Event Handler is a framework whose logic is implemented by users. It reacts 
+to the query results of  the Event Loop and generates SQL queries of the next 
+state. It stores the states in BucketFS.
+
+
+### Python UDF Framework 
+`dsn~python-udf-framework~1`
+
+An Event Handler designed as Python UDF  allows users to have an easy-to-use and 
+better tooling framework to build their algorithm on top it. The user can develop
```
MarleneKress79789, line 123:
```suggestion
An Event Handler designed as Python UDF instead of a Lua script allows users to have an easy-to-use and 
better tooling framework to build their algorithm on top of. The user can develop 
```

doc/design.md
```diff
@@ -0,0 +1,258 @@
...
+better tooling framework to build their algorithm on top it. The user can develop 
+the algorithm as a Python UDF script in which the framework is imported. Some 
+details of this framework interface are listed below:
+
+- The event handler is implemented as a class and states are kept as attributes 
+of a class to facilitate the states operations.
+- The user code should implement a method `handle_event` where the logic of the 
+code is placed.
+- The event handler is initiated by getting the following inputs: The results of 
+the return query,  the next state information including its parameters,
```
MarleneKress79789, line 132:
```suggestion
the return query, the next state information including its parameters, 
```

doc/design.md
```diff
@@ -0,0 +1,258 @@
...
+the algorithm as a Python UDF script in which the framework is imported. Some 
+details of this framework interface are listed below:
+
+- The event handler is implemented as a class and states are kept as attributes 
+of a class to facilitate the states operations.
+- The user code should implement a method `handle_event` where the logic of the 
+code is placed.
+- The event handler is initiated by getting the following inputs: The results of 
+the return query,  the next state information including its parameters, 
+context object.
```
MarleneKress79789, line 133: If you mean the ctx object in the udfs, not sure.
```suggestion
the ExasolDB context object. 
```

doc/design.md
```diff
@@ -0,0 +1,258 @@
...
+
+Covers:
+
+* `req~managing-temporary-bucketfs-files~1`
+
+Needs: impl, utest, itest
+
+
+It returns the sequence of SQL queries to execute
+A query to call the same UDF to run the action for the next state
```
MarleneKress79789, line 159:
```suggestion
a query to call the same UDF again to run the action for the next state.
```

doc/design.md
```diff
@@ -0,0 +1,258 @@
...
+
+* `feat~implementation-framework~1`
+
+Needs: impl, utest, itest
+
+### Managing Temporary BucketFS Files
+`dsn~managing-temporary-bucketfs-files~1`
+
+The Event Handler create temporary BucketFS files that will be  kept as result. 
+These temporary files are placed  in the same directory in  the BucketFS.
```
MarleneKress79789, line 178:
```suggestion
These temporary files are placed in the same directory in the BucketFS.   
```

doc/design.md
```diff
@@ -0,0 +1,258 @@
...
+
+* `req~electing-leader~1`
+
+Needs: impl, utest, itest
+
+
+### Collective Operation
+`dsn~collective-operation~1`
+
+The framework uses collective operation approach, where tasks are simultaneously
```
MarleneKress79789, line 219:
```suggestion
The framework uses a collective operation approach, where tasks are simultaneously 
```

doc/design.md
```diff
@@ -0,0 +1,258 @@
...
+
+### Error Handling
+In case of any errors in the execution, temporarily created files and tables 
+are cleaned. 
+
+
+#### Cleanup Temporary Tables
+`dsn~cleanup-temporary-tables~1`
+
+Temporary tables are removed in case of an error that occur during the execution
```
MarleneKress79789, line 237:
```suggestion
Temporary tables are removed in case of an error occuring during the execution 
```

doc/design.md
```diff
@@ -0,0 +1,258 @@
...
+### Error Handling
+In case of any errors in the execution, temporarily created files and tables 
+are cleaned. 
+
+
+#### Cleanup Temporary Tables
+`dsn~cleanup-temporary-tables~1`
+
+Temporary tables are removed in case of an error that occur during the execution 
+of the framework with the keeping option  not selected.
```
MarleneKress79789, line 238:
```suggestion
of the framework with the keeping option not selected.
```

doc/design.md
```diff
@@ -0,0 +1,258 @@
...
+* `req~electing-leader~1`
+
+Needs: impl, utest, itest
+
+
+### Collective Operation
+`dsn~collective-operation~1`
+
+The framework uses collective operation approach, where tasks are simultaneously 
+run on multiple UDF instances to achieve parallelism.
```
MarleneKress79789, line 220: Maybe expand on this a bit? (If you know more already.)

doc/design.md
```diff
@@ -0,0 +1,258 @@
...
+Covers:
+
+* `feat~implementation-framework~1`
+
+Needs: impl, utest, itest
+
+### Managing Temporary BucketFS Files
+`dsn~managing-temporary-bucketfs-files~1`
+
+The Event Handler create temporary BucketFS files that will be  kept as result.
```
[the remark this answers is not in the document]
umitbuyuksahin: when the execution of the framework is completed, we will no longer need these files. We might have finalization step where we cleanup them

doc/design.md
```diff
@@ -0,0 +1,258 @@
...
+of a class to facilitate the states operations.
+- The user code should implement a method `handle_event` where the logic of the 
+code is placed.
+- The event handler is initiated by getting the following inputs: The results of 
+the return query,  the next state information including its parameters, 
+context object. 
+- Upon completion of the algorithm, the Event Handler calls methods itself to 
+remove temporary records.
+- In case of an error caught in the Event Loop, the cleanup event is called for 
+the Event Handler and the temporary records are deleted.
```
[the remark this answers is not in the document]
umitbuyuksahin: You're right; I mentioned the keep option  in the requirements document, but overlooked it here.

MarleneKress79789 (approved):

doc/design.md
```diff
@@ -0,0 +1,258 @@
...
+
+## Assumptions
+
+- Since the number open result set handles in Lua scripts is limited, it is 
+assumed that Lua scripts is not nested more than 250 times.
+- The result of `pquery` is assumed to not exceed the db memory limit in size. 
+The DB usually allows 4 GB per query and node. The UDF can consume up to 32 GB 
+per node, but gets throttled down after the 4 GB. However, the 32 GB are shared 
+between all queries on a node.
+- The state of the algorithm is not too large.
```
tkilias, line 29:
```suggestion
- Because Python UDF can't run constantly, we need to save its state to the BucketFS. Saving and loading needs to reasonable fast to not introduce too much overhead. For that, we assume the state of the Python UDF is reasonable small (max a few MB).
```

tkilias (changes requested):

doc/design.md
```diff
@@ -0,0 +1,273 @@
...
+Below you can find the code snippet that simply indicates the Python framework 
+interface. Please note that the `return_query` is wrapped into the next call to 
+event handler UDF.
+
+```python
+class Result:
+  query_list: List[str]
+  return_query: Optional[str]
+
+def handle_event(row_iterator:RowIterable)->Result
```
tkilias, line 76:
```suggestion
class Context:
    """The Context is part of the Python framework and provides additional functions to the user code.
    
    def getTemporaryTableManager():TemporaryTableManager
    def getTemporaryBucketFSFileManager():TemporaryBucketFSFileManager
    

def handle_event(row_iterator:RowIterable,context:Context)->Result
```

doc/design.md
```diff
@@ -0,0 +1,279 @@
...
+class Context:
+    """The Context is part of the Python framework and provides additional functions to the user code."""
+    
+    def getTemporaryTableManager():TemporaryTableManager
+    def getTemporaryBucketFSFileManager():TemporaryBucketFSFileManager
+    
+def handle_event(row_iterator:RowIterable,context:Context)->Result
+    # user code
+```
+
```
tkilias, line 85: Move to `dsn~python-udf-framework~1,` I think there is a better fit

doc/design.md
```diff
@@ -0,0 +1,279 @@
...
+the algorithm as a Python UDF script in which the framework is imported. Some 
+details of this framework interface are listed below:
+
+- The event handler is implemented as a class and states are kept as attributes 
+of a class to facilitate the states operations.
+- The user code should implement a method `handle_event` where the logic of the 
+code is placed.
+- The event handler is initiated by getting the following inputs: The results of 
+the return query, the next state information including its parameters, 
+the ExasolDB context object.
```
tkilias, line 153:
```suggestion
- The first time the framework calls the event handler, it calls its constructor with a config __init__(config:Dict[str,Any]), the config is given to the event loop and the Python UDF in JSON format and deserialized to a Dict
- The event handler is called by getting the following inputs: The results of 
the return query, the ExasolDB context object. 
```

doc/design.md
```diff
@@ -0,0 +1,279 @@
...
+
+* `req~user-friendly-framework-interface~1`
+
+Needs: impl, utest, itest
+
+
+### Storing States in BucketFS 
+`dsn~storing-states-in-bucketfs~1`
+
+The framework keeps states during iterations by storing them in BucketFS.
```
tkilias, line 170:
```suggestion
The framework keeps states during iterations by storing them in BucketFS.
The states includes the event handler class of the user and the context state for managing temporary tables or bucketfs files
```

doc/design.md
```diff
@@ -0,0 +1,279 @@
...
+
+Covers:
+
+* `req~managing-temporary-bucketfs-files~1`
+
+Needs: impl, utest, itest
+
+
+It returns the sequence of SQL queries to execute
+a query to call the same UDF again to run the action for the next state.
```
tkilias, line 180: To which section does this belong to

doc/design.md
```diff
@@ -0,0 +1,279 @@
...
+These temporary files are placed in the same directory in the BucketFS.   
+
+Covers:
+
+* `req~managing-temporary-bucketfs-files~1`
+
+Needs: impl, utest, itest
+
+
+### Managing Temporary BucketFS Files
```
tkilias, line 208:
```suggestion
### Managing Temporary Tables
```

doc/system_requirements.md
```diff
@@ -0,0 +1,217 @@
+# System Requirement Specification Advanced Analytics Framework
```
tkilias, line 1: I guess, you need to merge the master into this branch, or?

tkilias (changes requested):

tkilias (approved):

PR #14, opened 2022-03-21
#7: Added event loop

### LINKED ISSUE #7: Add Lua Event Loop
# Background

- From work consists of an Event Loop and an Event-Handler
- The Event Loop calls the Event Handler and executes queries the Event Handler returned 
- The Event Loop needs to be written in Lua, such that we can run in a Lua Script to use pquery to run the queries
- The Event loop needs to receive configuration and must forward it to the event handler

# Acceptance Criteria

- Implement Event-Loop as Lua Module
- Implement Lua Script
- Implement deployment cli

umitbuyuksahin:
Fixes #7


exasol_advanced_analytics_framework/lua/src/event_loop.lua
```diff
@@ -0,0 +1,58 @@
...
+
+    repeat
+        local success, result = _G.global_env.pquery(query_to_event_handler)
+        if not success then
+            local error_obj = exa_error.create(
+                    "E-AAF-2",
+                    "Error occurred in calling Event Handler: " .. result.error_message)
+            _G.global_env.error(tostring(error_obj))
+        end
+        query_to_event_handler = result[1][1]
```
tkilias, line 49: So, I guess, this defines that the first result line of the event handler returns the return query.

exasol_advanced_analytics_framework/lua/src/event_loop_main.lua
```diff
@@ -0,0 +1,63 @@
+---
+-- @module event_loop_main
+--
+-- This scrip includes the main function of the Event Loop
```
tkilias, line 4:
```suggestion
-- This script includes the main function of the Event Loop
```

exasol_advanced_analytics_framework/lua/src/event_loop_main.lua
```diff
@@ -0,0 +1,63 @@
...
+---
+-- Prepare the initial query that initiates the Event Loop and calls Event Handler
+--
+function M._prepare_init_query(args)
+    local action = args['action'] -- TODO
+    local params = args['parameters']
+    local schema = args['schema']
+    local sql_prefix = args['sql_prefix']
+
+    local _udf_name = schema .. "." .. sql_prefix .."_EVENT_HANDLER_RUNNER_UDF"
```
tkilias, line 45: Maybe it is enough, when we specify which UDF should be called. This also means, the user can define its own UDFs which can be used with this event loop. Perhaps, also in different languages. As long, the UDF follows the behavioral contract, it doesn't matter. We can have of course have one predefined UDF which awaits an action (Python Class) in its parameter.

With this we only need the following arguments

parameters
schema
udf_name
bucketfs_connection

exasol_advanced_analytics_framework/lua/src/event_loop.lua
```diff
@@ -0,0 +1,58 @@
+---
```
tkilias, line 1: What do you think, can we integrate this module also in other repositories to build custom event loop scripts with maybe custom parameters?

tkilias (changes requested):

exasol_advanced_analytics_framework/lua/src/event_loop.lua
```diff
@@ -0,0 +1,58 @@
...
+    error = error
+}
+
+---
+-- Executes the given set of queries.
+--
+function M._run_queries(queries, from_index)
+    local all_success = true
+    for i=from_index, #queries do
+        local success, result = _G.global_env.pquery(queries[i][1])
```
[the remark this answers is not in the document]
umitbuyuksahin: "queries" is actually "result" of the query that calls Event Handler. I thought that EventHandlerUDF can return like that:

> ctx.emit(return_query)
> ctx.emit(status)
> ctx.emit(query_1)
> ctx.emit(query_2)
> ..
> ctx.emit(query_n)
> 

this sort of return creates 2-d matrix. Do you have better suggestion?

exasol_advanced_analytics_framework/lua/src/event_loop.lua
```diff
@@ -0,0 +1,58 @@
...
+        all_success = all_success and success
+        if not success then
+            local error_obj = exa_error.create(
+                    "E-AAF-3",
+                    "Error occurred in executing queries: " .. result.error_message)
+            _G.global_env.error(tostring(error_obj))
```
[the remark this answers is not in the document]
umitbuyuksahin: if you mean "error handling", I also thought it. But at this stage of the development it might be earlier to handle it. We can do it when the implementation is matured.

exasol_advanced_analytics_framework/lua/src/event_loop.lua
umitbuyuksahin: exactly as I wrote in the comment above.

exasol_advanced_analytics_framework/lua/src/event_loop_main.lua
```diff
@@ -0,0 +1,67 @@
...
+--
+-- @param json_str input parameters as json string
+--
+-- @return lua table including parameters
+--
+function M._parse_arguments(json_str)
+    local success, args =  pcall(json.decode, json_str)
+    if not success then
+		local error_obj = exaerror.create(
+                "E-SME-5",
```
tkilias, line 28: Wrong short tag

tkilias (changes requested):

tkilias (approved):

PR #21, opened 2022-04-12
#6: Added pseudo-code of the implementation

### LINKED ISSUE #6: Add Python Event-Handler Framework
# Background

- one of the parts of the Framework will be a python event handler. Usually, the logic of the event handler is implemented by the user, but we need to implement the Framework to execute this event handler.
- The Framework is responsible for the loading and storing the state, for call the user provided event handler and managing temporary tables
- To test event handler a python implementation of the event loop is useful, because debugging with the Lua event loop is probably difficult
- Users need to be able to specify the class which implements the event handler logic
  - Alternative, they might define a UDF from which this class is imported. This is for adhoc algorithms, when we don't want to build a python package or language container for the algorithm.

# Acceptance Criteria

- Create Python Event-Handler Framework
- Create Python Event Loop for Testing

umitbuyuksahin:
Fixes #6


exasol_advanced_analytics_framework/event_handler/event_handler_base.py
```diff
@@ -0,0 +1,19 @@
...
+from exasol_advanced_analytics_framework.event_handler.event_handler_result import \
+    EventHandlerResult
+
+
+class EventHandlerBase(ABC):
+    def __init__(self):
+        self.state = None
+
+    @abstractmethod
+    def handle_event(self) -> EventHandlerResult:
```
tkilias, line 11: here are some parameters missing, such as RowIterator and Context

exasol_advanced_analytics_framework/event_handler/event_handler_result.py
```diff
@@ -0,0 +1,9 @@
+from dataclasses import dataclass
+from typing import Optional, List
+
+
+@dataclass(unsafe_hash=True)
```
tkilias, line 5: what does this do?

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
```diff
@@ -1,10 +1,44 @@
...
 
     def run(self, ctx) -> None:
-        ctx.emit("return query")
-        ctx.emit("status")
-        ctx.emit("query 1")
-        ctx.emit("query 2")
+        parameters = ctx.parameters
+        bucketfs_connection = ctx.bucketfs_connection
+
+        event__handler_context = EventHandlerContext(bucketfs_connection)
```
tkilias, line 18: you need to handle here two cases, init (event handler doesn't exists, yet ), handle event (event handler exists, but needs to be deserialized)

exasol_advanced_analytics_framework/temporary_operations/temporary_bucketfs_file_manager.py
```diff
@@ -0,0 +1,19 @@
+
+
+class TemporaryBucketFSFileManager:
+    def __init__(self):
+        pass
+
+    def store_states(self):
```
tkilias, line 7: what do you understand under states, do you mean the user can directly store objects maybe with a key, this might be a useful functionality, but I would call it store/load object

tkilias (changes requested):

exasol_advanced_analytics_framework/temporary_operations/temporary_bucketfs_file_manager.py
umitbuyuksahin: agree, naming as load/store objects makes more sense

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
umitbuyuksahin: the method that is invoked just below (_get_event_handler_obj) actually does what you say.

exasol_advanced_analytics_framework/event_handler/event_handler_result.py
umitbuyuksahin: this is result object holds what the event handler should return

exasol_advanced_analytics_framework/event_handler/event_handler_base.py
umitbuyuksahin: Rowiterator, you mean the iterator from exa ctx, right?

tkilias: Yes, but we need to abstract it, in case we want to run the event handler somewhere else. Or, we keep the interface and build wrapper when we don't use UDFs for the event handler. However, you might need to filter out certain columns which are important for the framework, but the use shouldn't have access, too. I implemented something like that in the data-science-utils-python

exasol_advanced_analytics_framework/event_handler/event_handler_result.py
tkilias: I mean the unsafe hash

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
tkilias: Ah ok, the reason why I didn't recognize it, was that you also need to store and load the context it keeps the information about temporary objects

tkilias: My idea was, you have one object which keeps event handler and context as attributes, and you load and store this object

umitbuyuksahin: I had separated them. It's also a good idea to keep it together.

exasol_advanced_analytics_framework/event_handler/event_handler_result.py
umitbuyuksahin: aha it is not a crucial stuff. I did it like that because I thought the result object should be comparable and mutable. But comparability might be not necessary.

exasol_advanced_analytics_framework/temporary_operations/temporary_bucketfs_file_manager.py
```diff
@@ -1,19 +1,19 @@
...
-        pass
+    def save_object(self, object_to_load: Any, file_name: str) -> None:
+        self.bucketfs_location.upload_object_to_bucketfs_via_joblib(
+            object_to_load, str(PurePosixPath(self.bucketfs_path, file_name)))
 
-    def load_states(self):
-        pass
+    def load_object(self, file_name: str) -> Any:
+        self.bucketfs_location.download_object_from_bucketfs_via_joblib(
+            str(PurePosixPath(self.bucketfs_path, file_name)))
```
tkilias, line 18: Besides the functions we already have, we need the following `register_temporary_file` and `register_temporary_directory`  (this one doesn't create the file, but it registers one and return the location). Further, we probably need `save_fileobj` and `load_fileobj`.

exasol_advanced_analytics_framework/utils/udf_context_wrapper.py
```diff
@@ -0,0 +1,48 @@
+import pandas as pd
```
tkilias, line 1: move pandas import to get_dataframe, because we want to avoid, if we don't need it. It takes 0.5 seconds.

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
```diff
@@ -1,44 +1,93 @@
...
+from exasol_advanced_analytics_framework.event_handler.event_handler_result \
+    import EventHandlerResult
+from exasol_advanced_analytics_framework.event_handler.event_handler_state \
+    import EventHandlerState
+from exasol_advanced_analytics_framework.utils.udf_context_wrapper import \
+    UDFContextWrapper
 
 
 class CreateEventHandlerUDF:
+    STATE_DIRECTORY = "aaf_state"
```
tkilias, line 19: this one needs to be supplied via a connection object

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
```diff
@@ -1,44 +1,93 @@
...
 
-        event__handler_context = EventHandlerContext(bucketfs_connection)
-        event_handler_obj = self._get_event_handler_obj(
-            parameters,  event__handler_context)
-        result: EventHandlerResult = event_handler_obj.handle_event()  # TODO ?
-        self._save_event_handler_obj(event_handler_obj, event__handler_context)
+        # call the user code
+        udf_context = self._create_udf_context_wrapper(ctx)
+        result: EventHandlerResult = event_handler.handle_event(
+            udf_context, event_handler_context)  # TODO does user change state?
```
tkilias, line 46: the attributes of the `event_handler` and `event_handler_context` will be changed after calling `handle_event`

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
```diff
@@ -1,44 +1,93 @@
...
 
+        # update and save (overwrite)the current state
+        current_state = EventHandlerState(event_handler_context, event_handler)
+        self._save_current_state(
```
tkilias, line 50: maybe store new state first to a new file and then remove the old one, such that we have a safe checkpoint

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
```diff
@@ -1,44 +1,93 @@
...
 
-    def _get_event_handler_obj(
-            self, parameters, event__handler_context) -> EventHandlerBase:
+    def _load_latest_state(
+            self,
+            parameters: dict,
+            bucketfs_location: BucketFSLocation,
+            bucketfs_path: PurePosixPath) -> EventHandlerState:
+
+        event_handler_state = bucketfs_location.\
```
tkilias, line 65: This will throw an exception if the file is not available. I suggest, you check if the file is actually available. Maybe, it is also better, to explicitly inform the UDF about which state to load or if it gets initialized.

exasol_advanced_analytics_framework/event_handler/event_handler_base.py
```diff
@@ -1,18 +1,23 @@
...
+from exasol_advanced_analytics_framework.event_handler.event_handler_context \
+    import EventHandlerContext
+from exasol_advanced_analytics_framework.event_handler.event_handler_result \
+    import EventHandlerResult
+from exasol_advanced_analytics_framework.utils.udf_context_wrapper \
+    import UDFContextWrapper
 
 
 class EventHandlerBase(ABC):
     def __init__(self):
```
tkilias, line 11: The event handler gets a Python Dict as input for initialization

tkilias (changes requested):

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
umitbuyuksahin: since I use same location and path, isn't it overwriting?

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
umitbuyuksahin: not sure about the way of explicitly informing UDF,  since we have to manipulate sql query where we invoke  in event loop (saying that it is the first call). did you mean that too?

I guess we can't check whether the file exists with the bucketfs_location api. Exception handling is not a good way.

exasol_advanced_analytics_framework/temporary_operations/temporary_bucketfs_file_manager.py
umitbuyuksahin: what does registering to a file ('register_temporary_file' and 'register_temporary_directory') meaning?

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
tkilias: yes it is overwriting and this is a problem. To keep a safe checkpoint, you need to create a new file and then delete the old one.

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
tkilias: We might need to extend the bucketfs utils for that. Yes, I mean that event loop could inform the UDF that this is the first call. You could use different mechanism for that. Either you UDF parameter or use a connection object. For example, you could add a column to the UDF input, which contains either the location of the checkpoint or the config.

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
umitbuyuksahin: then we can do something like this that will also solve the following problem:

-  when calling udf query on event-loop side, we can add a parameter like step_num to udf query
- If step_num = 0, we know that there is no saved state yet. So, create new one.
- Adding step_num to the name of state objects makes state objs unique (remove state_obj_[step_num-1], save state_obj_[step_num])

tkilias: Sounds good. For step num 0 we can also add the config as a parameter to the UDF. On that note, I remember something. The UDF for the event handler needs to have dynamic input parameter.

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
```diff
@@ -1,10 +1,93 @@
...
+        # get and set method parameters
+        parameters = ctx.parameters
+        bucketfs_connection = ctx.bucketfs_connection
+        bucketfs_location = BucketFSFactory().create_bucketfs_location(
+            url=bucketfs_connection.address,
+            user=bucketfs_connection.user,
+            pwd=bucketfs_connection.password,
+            base_path=parameters["base"]
+        )
+        bucketfs_path = PurePosixPath(
```
tkilias, line 33: I think, we don't need this, we can manage everything via the bucketfs_connection and the step parameter

exasol_advanced_analytics_framework/lua/src/event_loop.lua
```diff
@@ -41,20 +86,34 @@ end
 ---
 -- Initiates the Event Loop that handles state transition
 --
--- @param query string that calls the event handler
+-- @param schema        name of schema where the Event-Handler udf script is installed
+-- @param class_name    name of the class built on the Event-Handler
+-- @param bucketfs_conn the name of BucketFS connection
+-- @parameters           config parameters of the Event-Handler
```
tkilias, line 92: rename to event_hanlder_parameters

tkilias (changes requested):

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
```diff
@@ -22,58 +24,85 @@ def __init__(self, exa):
...
+            # load the latest state
+            event_handler_state = bucketfs_location.\
+                download_object_from_bucketfs_via_joblib(str(bucketfs_path))
+        else:
+            # create the new state
             context = EventHandlerContext(
                 bucketfs_location, bucketfs_path)
             event_handler_class = getattr(importlib.import_module(
-                parameters["module_name"]), parameters["class_name"])
+                'module_name'), event_handler_class)  # TODO-1: module name how to get
```
umitbuyuksahin, line 93: we also need to get 'module_name' from user together with 'class_name', what do you think?

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
```diff
@@ -22,58 +24,85 @@ def __init__(self, exa):
...
             event_handler_obj = event_handler_class()
             event_handler_state = EventHandlerState(context, event_handler_obj)
 
         return event_handler_state
 
+    @staticmethod
+    def _remove_previous_state(
+            bucketfs_location: BucketFSLocation,
+            bucketfs_path: PurePosixPath) -> None:
+        # TODO-2: requests.delete(url, ..)
```
umitbuyuksahin, line 103: bucketfs-utils does not provide any 'delete'  method. Should we send an explicit delete request to Bucketfs or is there any better way to do it?

exasol_advanced_analytics_framework/event_handler/event_handler_result.py
```diff
@@ -1,9 +1,10 @@
 from dataclasses import dataclass
-from typing import Optional, List
+from typing import Optional, List, Dict, Any
 
 
 @dataclass()
 class EventHandlerResult:
     return_query: Optional[str]
+    return_query_columns: Optional[Dict[str, Any]]
```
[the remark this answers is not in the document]
tkilias: yes, this is what I also thought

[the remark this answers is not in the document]
tkilias: Maybe create a new class for the return  query where you both encapsulates, Also, a question for what is the Dict in the columns.

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
tkilias: yes, you are right, we need both. The question is, do we want it as <module_path>.<class_name> or separate.

exasol_advanced_analytics_framework/event_handler/event_handler_result.py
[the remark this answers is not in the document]
tkilias: For the table definition of return query, you might want to use the following classes

https://github.com/exasol/data-science-utils-python/tree/main/exasol_data_science_utils_python%2Fpreprocessing%2Fsql%2Fschema

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
```diff
@@ -22,58 +24,85 @@ def __init__(self, exa):
...
+            bucketfs_location, latest_bucketfs_path)
+
+        # wrap return query
+        return_query_view, return_query = self._wrap_return_query(
+            iter_num, bucketfs_connection,
+            result.return_query, result.return_query_columns)
 
         # return queries
-        ctx.emit(result.return_query)
+        ctx.emit(return_query_view)  # TODO-4
```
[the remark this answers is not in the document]
tkilias: Looks good

tkilias (changes requested):
See my comments and I don't see how you execute the return query view, maybe event_loop.lua needs an update.

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
umitbuyuksahin: yes, this was what I wanted to ask for. I think getting them separately is better

exasol_advanced_analytics_framework/event_handler/event_handler_result.py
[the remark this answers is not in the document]
umitbuyuksahin: - Dictionary includes column names and column types. Column types might be redundant

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
```diff
@@ -1,10 +1,135 @@
...
+        column_mapping = OrderedDict([
+            (str(4 + index), column)
+            for index, column in enumerate(column_name_list)])
+        return UDFContextWrapper(ctx, column_mapping=column_mapping)
+
+    @staticmethod
+    def _wrap_return_query(
+            iter_num: int, bucketfs_conn: str,
+            query: str, query_columns: dict) -> Tuple[str,str]:
+        columns_str = ",".join(query_columns.keys())
```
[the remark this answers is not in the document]
umitbuyuksahin: As far as I see, even if I use those classes in data_science_utils_python , I need to handle sql statement creation using string operations and handling quotation.

exasol_advanced_analytics_framework/lua/src/event_loop.lua
```diff
@@ -45,18 +45,17 @@ end
...
+        local result = M._run_queries(return_queries, 1)
+
+        -- handle EventHandlerUDF return
+        query_to_create_view = result[1][1]
+        query_to_event_handler = result[2][1]
+        status = result[3][1]
+        M._run_queries(result, 4)
     until (status == 'completed')
 
     return status -- TODO return
```
tkilias, line 61: this is a good point, when we get the status completed, the event handler should have the option to return a json object (Dict[str, any])

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
```diff
@@ -126,13 +125,19 @@ def _create_udf_context_wrapper(ctx) -> UDFContextWrapper:
...
-        tmp_view_name = "tmp_view".upper()
-        event_handler_udf_name = "AAF_EVENT_HANDLER_UDF".upper()
-        query_create_view = f"Create view {tmp_view_name} as {query};"
-        query_event_handler = f"SELECT {event_handler_udf_name}" \
-                              f"({iter_num},'{bucketfs_conn}',{columns_str}) " \
-                              f"FROM {tmp_view_name};"
+            iter_num: int, bucketfs_conn: str, schema: str,
+            return_query: EventHandlerReturnQuery) \
+            -> Tuple[str, str]:
+        tmp_view_name = \
```
tkilias, line 131: Please, use these classes, they can handle quoting and also validate for valid identifier

https://github.com/exasol/data-science-utils-python/tree/main/exasol_data_science_utils_python/preprocessing/sql/schema

exasol_advanced_analytics_framework/event_handler/event_handler_result.py
```diff
@@ -2,9 +2,14 @@
 from typing import Optional, List, Dict, Any
 
 
+@dataclass()
+class EventHandlerReturnQuery:
+    query: str
+    query_columns: List[str]
```
tkilias, line 8: Please, use these classes, they can handle quoting and also validate for valid identifier

https://github.com/exasol/data-science-utils-python/tree/main/exasol_data_science_utils_python/preprocessing/sql/schema

tkilias (changes requested):
We are getting closer. Here some more comments.

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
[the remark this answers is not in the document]
tkilias: These classes handle quoting and validation. And using object passing around column names is much more expressive. So, please use them, trust me, I was already at that point, we didn't have them and it got too much to handle. Long term they really improve the code base and not only for use, also for the implementer of the algorithms

exasol_advanced_analytics_framework/event_handler/event_handler_result.py
```diff
@@ -11,7 +11,16 @@ class EventHandlerReturnQuery:
 
 
 @dataclass()
-class EventHandlerResult:
-    return_query: Optional[EventHandlerReturnQuery]
+class EventHandlerResultBase:
     status: str
```
tkilias, line 15: I think, we don't need the status attribute anymore. We can derive it from the Class Type of the result object. This is also safer, because the event loop expects specific status values

exasol_advanced_analytics_framework/event_handler/event_handler_result.py
```diff
@@ -11,7 +11,16 @@ class EventHandlerReturnQuery:
...
     query_list: List[str]
```
tkilias, line 16: The EventHandlerResultFinished doesn't need a query_list, we only create a query_list in the Python UDF to cleanup the temporary tables

tkilias (changes requested):

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
```diff
@@ -71,19 +71,23 @@ def run(self, ctx) -> None:
         return_query_view = None
         return_query = None
         final_result = {}
-        if result.status != "finished":
-            schema = self.exa.meta.script_schema
-            return_query_view, return_query = self._wrap_return_query(
-                iter_num, bucketfs_connection, schema, result.return_query)
-        else:
+        query_list = []
+        if result.is_finished:
```
tkilias, line 75: Here it is probably better to check the type via instance of, otherwise the mypy later throws errors when access the attributes of the subclass

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
```diff
@@ -1,10 +1,162 @@
...
+        iter_num += 1
+        current_bucketfs_path = PurePosixPath(
+            f"{event_handler_class}_{str(iter_num)}.pkl")
+        current_state = EventHandlerState(event_handler_context, event_handler)
+        self._save_current_state(
+            bucketfs_location, current_bucketfs_path, current_state)
+
+        # remove previous state
+        self._remove_previous_state(
+            bucketfs_location, latest_bucketfs_path)
```
tkilias, line 68: refactor: extract to method

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
```diff
@@ -1,10 +1,162 @@
...
+        # call the user code
+        udf_context = self._create_udf_context_wrapper(ctx)
+        result: EventHandlerResultBase = event_handler.handle_event(
+            udf_context, event_handler_context)
+
+        # save the current state
+        iter_num += 1
+        current_bucketfs_path = PurePosixPath(
+            f"{event_handler_class}_{str(iter_num)}.pkl")
+        current_state = EventHandlerState(event_handler_context, event_handler)
```
tkilias, line 62: you need to store the columns of the return_query, such that you can use them in the next iteration for creating the UDFContext

exasol_advanced_analytics_framework/context_wrapper/context_wrapper_base.py
```diff
@@ -0,0 +1,7 @@
+from abc import ABC
+
+
+class ContextWrapperBase(ABC):
```
tkilias, line 4: We need to rename the ContextWrapper, because it should be an abstraction for the results of the return query. We want to hide the UDF specifics, because we might want to implement the EventLoop differently in the future, so we should reduce the interface to the minimum necessary and as such also adapt its name. I suggest, we call it EventContext, because it contains the Context for the current Event. I would suggest it defines the same methods as the ResultSet in https://github.com/exasol/data-science-utils-python/blob/45283be1e0a61894249cef2d2bbb4e4c4c16b2fb/exasol_data_science_utils_python/udf_utils/sql_executor.py#L5. The ResultSet doesn't provide a reset, because this is usually not supported by our clients and the UDFs are one of the rare examples, where we support this.

tkilias (changes requested):

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
umitbuyuksahin: maybe it's better to remove class inheritance, it doesn't seem necessary after our last update.

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
umitbuyuksahin: I didn't understand the refactoring you mean

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
umitbuyuksahin: In addition, even_handler_context is already part of the event_handler object. It will be redundant to keep in state object.

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
tkilias: yes, but the check of type is still necessary for mypi

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
tkilias: Extract the storing new state and removing old state into a method.

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
tkilias: But event_handler_context doesn't store the columns of the return query and no one else does it, that means you need to add them EventHandlerState. I think that is also where they belong to. They are not part of the event_handler_context, because columns of the return query are only for the next iteration and the event_handler_context exists over the whole lifetime of the event handler.

umitbuyuksahin: No, you misunderstood me. We have to add retrun_query columns, there is no problem with that. I will update what I mean in the next PR

tkilias: ok, I now understand your sentence, but I disagree, the event_handler_context is not part of the event_handler, we only give it via the handle_next. We could think, about injecting via the constructor of the event_hanlder. This, however, has the risks, because we can't replace it through the iterations, if the implementation would make it necessary.

exasol_advanced_analytics_framework/event_context/udf_event_context.py
```diff
@@ -1,10 +1,18 @@
...
+from exasol_data_science_utils_python.preprocessing.sql.schema.column_type import \
+    ColumnType
 
-class UDFContextWrapper(ContextWrapperBase):
+from exasol_advanced_analytics_framework.event_context.event_context_base \
+    import EventContextBase
+
+
+class UDFEventContext(EventContextBase):
     def __init__(self, ctx, column_mapping: Mapping[str, str],
```
tkilias, line 16: We need to inject the exa object via the constructor for getting the column types

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
```diff
@@ -141,33 +139,23 @@ def _handle_states(
             event_handler: EventHandlerBase,
             current_bucketfs_path: PurePosixPath,
             latest_bucketfs_path: PurePosixPath,
+            query_columns: List[Column],
```
tkilias, line 142: I think, we don't need this anymore.

tkilias (changes requested):

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
```diff
@@ -32,30 +36,28 @@ def __init__(self, exa):
...
         event_handler_class = ctx[2]  # event_handler_class_name
-        bucketfs_connection = ctx[3]  # bucketfs_connection_name
-        parameters = ctx[4]  # event_handler_parameters
 
         bucketfs_location = BucketFSFactory().create_bucketfs_location(
             url=bucketfs_connection.address,
             user=bucketfs_connection.user,
-            pwd=bucketfs_connection.password,
-            base_path=parameters["base"]
+            pwd=bucketfs_connection.password
```
tkilias, line 45: bucketfs_connection is only the connection name, you still need to fetch the connection object via exa.get_connection

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
```diff
@@ -1,10 +1,194 @@
...
+        # call the user code
+        udf_event_context = self._create_udf_event_context(
+            ctx, iter_num, query_columns)
+        result: EventHandlerResultBase = event_handler.handle_event(
+            udf_event_context, event_handler_context)
+
+        # save the current state and remove the previous state
+        iter_num += 1
+        current_bucketfs_path = PurePosixPath(
+            f"{event_handler_class}_{str(iter_num)}.pkl")
```
tkilias, line 67: should we maybe use the fully qualified name of the event handler, meaning with the module name

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
```diff
@@ -1,10 +1,194 @@
...
+    def _handle_states(
+            self,
+            event_handler_context: EventHandlerContext,
+            event_handler: EventHandlerBase,
+            current_bucketfs_path: PurePosixPath,
+            latest_bucketfs_path: PurePosixPath,
+            bucketfs_location: BucketFSLocation) -> None:
+
+        current_state = EventHandlerState(
+            event_handler_context, event_handler, self.get_query_columns())
```
tkilias, line 149: I think, using self.get_query_columns() doesn't make sense, because self.get_query_columns() create the column from the exa.meta object. The names their are simply the numbers of the columns, because we use a Variadic UDF. I think, here you want to use the columns of the return column and function get_query_columns together with columns of the return query of the previous iteration when creating the EventContext.

tkilias (changes requested):

exasol_advanced_analytics_framework/interface/create_event_handler_udf.py
umitbuyuksahin: in the iterations after the first iteration, we need event_class_name to read previous state. Because we saved states with the event_class_name.  If we also use event_module_name while saving , we have to add event_module_name to UDF query at each iteration. Do we want to do that?

tests/udf_framework/test_create_event_handler_udf_mock.py
```diff
@@ -0,0 +1,143 @@
...
+    from exasol_udf_mock_python.udf_context import UDFContext
+    from exasol_advanced_analytics_framework.udf_framework.\
+        create_event_handler_udf import CreateEventHandlerUDF
+
+    udf = CreateEventHandlerUDF(exa)
+    def run(ctx: UDFContext):
+        udf.run(ctx)
+
+
+def _create_mock_data():
```
tkilias, line 24: create_mock_meta_data

tests/udf_framework/test_create_event_handler_udf_mock.py
```diff
@@ -0,0 +1,143 @@
...
+            0,
+            "bucketfs_connection",
+            "MockEventHandlerWithOneIteration",
+            "tests.udf_framework.mock_event_handlers",
+            "{}"
+        )
+        result = executor.run([Group([input_data])], exa)
+        for i, group in enumerate(result):
+            result_row = group.rows
+            assert len(result_row) == 4
```
tkilias, line 64: combine asserts

tests/udf_framework/test_create_event_handler_udf_mock.py
```diff
@@ -0,0 +1,143 @@
...
+            0,
+            "bucketfs_connection",
+            "MockEventHandlerWithTwoIterations",
+            "tests.udf_framework.mock_event_handlers",
+            "{}"
+        )
+        result = executor.run([Group([input_data])], exa)
+        for i, group in enumerate(result):
+            result_row = group.rows
+            assert len(result_row) == 4 + len(mock_event_handlers.QUERY_LIST)
```
tkilias, line 91: use for the multiple assert pytest-assume https://github.com/astraw38/pytest-assume

tkilias (changes requested):
see my comments. I think, we should add an integration test to also test the event loop and then merge this PR. And create new ticket and PR to complete the missing parts, because this PR is already huge.

tests/event_context/test_udf_event_context.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def run(ctx: UDFContext):
+        wrapper = UDFEventContext(
+            ctx, exa, OrderedDict([("t2", "a"), ("t1", "b")]))
+        df = wrapper.fetch_as_dataframe(10)
+        assert len(df) == 10
+        assert list(df.columns) == ["a", "b"]
+        assert wrapper.column_names() == ["t1", "t2", "t3"]
+        assert wrapper.rowcount() == 100
+
```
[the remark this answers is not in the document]
umitbuyuksahin: but the wrapper is only available inside of UDF. That's why I didn't move the regarding asserts. I moved the others.

tests/udf_framework/mock_event_handlers.py
```diff
@@ -0,0 +1,53 @@
...
+        return EventHandlerResultFinished(final_result=FINAL_RESULT)
+
+
+class MockEventHandlerWithTwoIterations(EventHandlerBase):
+    def handle_event(self,
+                     exa_context: EventContextBase,
+                     event_handler_context: EventHandlerContext) -> \
+            EventHandlerResultBase:
+
+        is_last_iteration = exa_context.ctx[0] > 0
```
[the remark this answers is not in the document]
umitbuyuksahin: This is a control placed to make the test (with two iteration) flow possible.

The iteration number is already tracked  and changed in UDF.

tests/udf_framework/test_create_event_handler_udf_mock.py
umitbuyuksahin: why do you want to combine them? I did it but I didnt get the point

tkilias: Pytest doesn't really work with multiple asserts. The first assert which hits throws an exception and everything else is not evaluated. There are cases where this is fine, but in others not

  https://www.goshpat.com/multiple-assert-statements.html

tests/udf_framework/mock_event_handlers.py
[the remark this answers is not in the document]
tkilias: If the serialization works, you can have a internal counter in the EventHandler.

tests/event_context/test_udf_event_context.py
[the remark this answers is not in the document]
umitbuyuksahin: used json serialization , which was easy.

tests/event_context/test_udf_event_context.py
```diff
@@ -16,36 +16,45 @@ def udf_wrapper():
...
     from collections import OrderedDict
+    import json
 
     def run(ctx: UDFContext):
         wrapper = UDFEventContext(
             ctx, exa, OrderedDict([("t1", "a"), ("t3", "b"), ("t2", "c")]))
         df = wrapper.fetch_as_dataframe(10)
-        assert wrapper.column_names() == ["t1", "t2", "t3"]
-        assert wrapper.rowcount() == 100
+        return_str = json.dumps(str({
```
tkilias, line 25: The longer , I think about this test, the more I feel, we lag a something in the UDF mock, maybe we should provide a context implementation without the need to use the executor. I think, I will add a ticket for that.

exasol_advanced_analytics_framework/event_context/event_context_base.py
```diff
@@ -0,0 +1,30 @@
+from abc import ABC, abstractmethod
+from typing import Union, List
+from exasol_data_science_utils_python.preprocessing.sql.schema.column import \
+    Column
+
+
+class EventContextBase(ABC):
+    def __init__(self, ctx):
+        self.__ctx = ctx
```
tkilias, line 9: While I see it, the base doesn't know the ctx object , only the implementation UDFEventContext for the UDFs knows it. The reason is, that we might want to replace the UDFEventContext

tests/udf_framework/mock_event_handlers.py
```diff
@@ -0,0 +1,59 @@
...
+            return_query_columns = [
+                Column(ColumnName("a"), ColumnType("INTEGER")),
+                Column(ColumnName("b"), ColumnType("INTEGER"))]
+            event_handler_return_query = EventHandlerReturnQuery(
+                query=return_query,
+                query_columns=return_query_columns)
+            event_handler_result = EventHandlerResultContinue(
+                query_list=QUERY_LIST,
+                return_query=event_handler_return_query)
+            self.iter += 1
```
tkilias, line 58: The increment is not part of the else

tkilias (changes requested):

tests/udf_framework/test_create_event_handler_udf_mock.py
```diff
@@ -88,21 +89,21 @@ def test_event_handler_udf_with_two_iteration():
...
+            assert len(result_row) == 4 + len(mock_event_handlers.QUERY_LIST) \
+                   and is_finished == "False" \
+                   and query_view == "Create view \"TEST_SCHEMA\".\"TMP_VIEW\" " \
+                                     "as SELECT a, table1.b, c FROM table1, " \
+                                     "table2 WHERE table1.b=table2.b;" \
+                   and query_return == "SELECT \"TEST_SCHEMA\"." \
+                                       "\"AAF_EVENT_HANDLER_UDF\"(1," \
+                                       "'bucketfs_connection',\"a\",\"b\") " \
+                                       "FROM \"TEST_SCHEMA\".\"TMP_VIEW\";" \
+                   and pytest.assume(
```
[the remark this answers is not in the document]
umitbuyuksahin: seems outdated comment.

tests/udf_framework/test_create_event_handler_udf_mock.py
```diff
@@ -0,0 +1,144 @@
...
+            "MockEventHandlerWithTwoIterations",
+            "",
+            ""
+        )
+        result = executor.run([Group([input_data])], exa)
+        for i, group in enumerate(result):
+            result_row = group.rows
+            is_finished = result_row[2][0]
+            final_result = result_row[3][0]
+            assert len(result_row) == 4
```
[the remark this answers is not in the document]
umitbuyuksahin: seems outdated comment.

tests/udf_framework/test_create_event_handler_udf_mock.py
```diff
@@ -103,12 +103,10 @@ def test_event_handler_udf_with_two_iteration():
                    and set(mock_event_handlers.QUERY_LIST) == set(
                             list(map(lambda x: x[0], result_row[4+i:])))
 
-
-        # Comment out due to the ticket #66 - Correct listing files method for local operations
-        # assert _is_state_exist(
-        #     1, "MockEventHandlerWithTwoIterations", bucketfs_connection)
-        # assert not _is_state_exist(
-        #     0, "MockEventHandlerWithTwoIterations", bucketfs_connection)
+        assert _is_state_exist(1, "MockEventHandlerWithTwoIterations",
```
tkilias, line 106: I think, here, it would be best to compute first both function calls and store the result and variables and then merge the assert with the previous assert

tkilias (changes requested):

tkilias (approved):

PR #22, opened 2022-04-19
#20: Saved the bundled lua script

### LINKED ISSUE #20: Store bundled Lua scripts as resource
- in the current installation, the bundled Lua script is generated while installing the project
- however this installation requires Lua and its dependencies to be installed in the user machine.
- in order to break this dependencies, we should store the bundled Lua scripts in the wheel.

- [ ] add functionality to save bundled script
- [ ] provide an option to cli whether to re-bundle  or not
- [ ] call re-bundling in each commit to be sure that it is saved

umitbuyuksahin:
Fixes #20


exasol_advanced_analytics_framework/deployment/scripts_deployer_cli.py
```diff
@@ -10,8 +10,10 @@
 @click.option('--pass', 'pwd', type=str)
 @click.option('--schema', type=str, required=True)
 @click.option('--language-alias', type=str, default="PYTHON3_AAF")
-def scripts_deployer_main(dsn: str, user: str, pwd: str,
-                          schema: str, language_alias: str):
+@click.option('--develop', type=bool, is_flag=True, required=False)
```
Nicoretti, line 13: options by default should not be required, required options usually tend to indicate that the CLI interface could use some rework.

exasol_advanced_analytics_framework/resources/outputs/create_event_loop.sql
```diff
@@ -0,0 +1,564 @@
+CREATE OR REPLACE LUA SCRIPT "AAF_EVENT_LOOP"(json_str)  AS
+    table.insert(_G.package.searchers,
+        function (module_name)
+            local loader = package.preload[module_name]
+            if not loader then
+                error("Module " .. module_name .. " not found in package.preload.")
+            else
+                return loader
+            end
```
Nicoretti, line 9: The else case tends to add unnecessary extra noise especially in langauges where the blocks (syntax) itself produce significant noise (e.g. lua and ruby).

Imho it has a very bad signal to noise ratio here:
Signal:
* `return loader`
Noise:
* 1 additional block (branch)
* 1 additional keyword `else`

if you get rid of the additional branch, which will be executed in all non error cases it is way easier to read (imho)
```suggestion
           end
           return loader
```

exasol_advanced_analytics_framework/resources/outputs/create_event_loop.sql
```diff
@@ -0,0 +1,564 @@
...
+-- @see M:new
+--
+-- @raise Lua error for the given error object
+--
+function M.error(arg1, ...)
+    if type(arg1) == "table" then
+        M:new(arg1):raise()
+    else
+        M.create(arg1, ...):raise()
+    end
```
Nicoretti, line 302:
```suggestion
    end
    M.create(arg1, ...):raise()
```

exasol_advanced_analytics_framework/resources/outputs/create_event_loop.sql
```diff
@@ -0,0 +1,564 @@
...
+-- </p>
+--
+-- @return expanded message
+--
+function M:expand()
+    if (self.parameters == nil) or (not next(self.parameters)) then
+        return self.message
+    else
+        self:run_()
+    end
```
Nicoretti, line 372:
```suggestion
    end
    self:run_()
```

exasol_advanced_analytics_framework/resources/outputs/create_event_loop.sql
```diff
@@ -0,0 +1,564 @@
...
+    if parameter == nil then
+        return "missing value"
+    else
+        local type = type(parameter)
+        if (type == "table") then
+            return parameter.value
+        else
+            return parameter
+        end
+    end
```
Nicoretti, line 450:
```suggestion
    end
    local type = type(parameter)
    if (type == "table") then
        return parameter.value
    end
    return parameter
```

exasol_advanced_analytics_framework/deployment/bundle_lua_scripts.py
```diff
@@ -54,3 +57,24 @@ def get_content(cls) -> str:
...
+            autoescape=select_autoescape()
+        )
+        template = env.get_template(constants.LUA_SCRIPT_TEMPLATE)
+        lua_query = template.render(
+            bundled_script=lua_bundled_content)
+        return lua_query
+
+    @staticmethod
+    def save_statement() -> None:
+        stmt = BundleLuaScripts.get_statement()
```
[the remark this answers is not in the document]
umitbuyuksahin: This class and the methods in it are not generic,  the statement always indicates the same thing

Nicoretti (approved):

PR #27, opened 2022-08-10
#28: Extended the EventHandlerContext to a scope-based system for handling temporary objects 

### LINKED ISSUE #28: Extend the EventHandlerContext to a scope-based system for handling temporary objects
# Background

- Currently, the EventHandlerContext only provides an interface to create and load temporary bucketfs files, but didn't have yet a way to clean up
- Implementing the EventHandlerContext as a scope-based system gives us finer control over when temporary objects get cleaned up. This is important, in case we execute a long sequence of operations, where each operation can create temporary objects. At some point, some temporary objects are not needed anymore and can be release. Without that, we would potentially accumulate many large tables.
- In the scope-based EventHandlerContext, a parent event handler creates a ScopeEventHandlerContext and forwards it to a child event handler. The child event handler uses the EventHandlerContext for creating temporary objects. If the child event handler finishes, the parent releases the ScopeEventHandlerContext and with that, all temporary objects created by the child event handler.
- We can, of course, build the scopes recursively, where all intermediate event handlers get ScopeEventHandlerContext from which they can create new ChildEventHandlerContext.
- On the top layer we have a special EventHandlerContext which finally can clean up the release objects after the event handler returned
- Furthermore, we need to be able to transfer temporary objects from one scope to another scope, in case an event handler returns it or gets it as input. For this, a temporary objects needs to be owned by exactly one ScopeEventHandlerContext and we need a way to transfer ownership between the ScopeEventHandlerContext (parent, child, sibling)
- Also, if a ScopeEventHandlerContext forgets to release its temporary objects, they should get released by its parent ScopeEventHandlerContext. This ensures, at least at the end of the event_handler we clean up.

tkilias:
Fixes #28


exasol_advanced_analytics_framework/event_handler/context/top_level_event_handler_context.py
```diff
@@ -0,0 +1,219 @@
...
+        self._valid_object_proxies: Set[ObjectProxy] = set()
+        self._invalid_object_proxies: Set[ObjectProxy] = set()
+        self._owned_object_proxies: Set[ObjectProxy] = set()
+        self.__counter = 0
+        self._child_event_handler_context_list: List[_ChildEventHandlerContext] = []
+        self._is_valid = True
+
+    def release(self):
+        self._check_if_valid()
+        for object_proxy in list(self._valid_object_proxies):
```
umitbuyuksahin, line 36: not need to cast the set object to list

exasol_advanced_analytics_framework/event_handler/context/top_level_event_handler_context.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def _release_object(self, object_proxy: ObjectProxy):
+        self._check_if_valid()
+        self._valid_object_proxies.remove(object_proxy)
+        if object_proxy in self._owned_object_proxies:
+            self._owned_object_proxies.remove(object_proxy)
+        self._invalid_object_proxies.add(object_proxy)
+        object_proxy._invalidate()
+
+    def _register_object(self, object_proxy: ObjectProxy):
```
umitbuyuksahin, line 149: I would implement the  _register_object() method in parent class, because the task of adding the object into valid_proxies should be same for all subclasses.

exasol_advanced_analytics_framework/event_handler/context/top_level_event_handler_context.py
```diff
@@ -0,0 +1,219 @@
...
+    def _parent(self) -> ScopeEventHandlerContext:
+        return self.__parent
+
+    def _release_object(self, object_proxy: ObjectProxy):
+        self._check_if_valid()
+        self._valid_object_proxies.remove(object_proxy)
+        self._invalid_object_proxies.add(object_proxy)
+        self._parent._release_object(object_proxy)
+
+    def _register_object(self, object_proxy: ObjectProxy):
```
umitbuyuksahin, line 198: In case of implementing the register_object() in parent class, we should change here
```
super().register_object(object)
self._parent._register_object(object_proxy)
```

exasol_advanced_analytics_framework/event_handler/context/top_level_event_handler_context.py
```diff
@@ -0,0 +1,219 @@
...
+        child_event_handler_conext = _ChildEventHandlerContext(
+            self,
+            new_temporary_bucketfs_location,
+            self._get_temporary_db_object_name(),
+            self._temporary_schema_name
+        )
+        self._child_event_handler_context_list.append(child_event_handler_conext)
+        return child_event_handler_conext
+
+    def _is_child(self, scope_event_handler_context: ScopeEventHandlerContext):
```
umitbuyuksahin, line 99:
```suggestion
    def _is_child(self, scope_event_handler_context: ScopeEventHandlerContext) -> bool:
```

exasol_advanced_analytics_framework/event_handler/context/top_level_event_handler_context.py
```diff
@@ -0,0 +1,219 @@
...
+        result = isinstance(scope_event_handler_context, _ChildEventHandlerContext) and \
+                 scope_event_handler_context._parent == self
+        return result
+
+    def _transfer_object_to(self, object_proxy: ObjectProxy,
+                            scope_event_handler_conext: ScopeEventHandlerContext):
```
umitbuyuksahin, line 105:
```suggestion
                            scope_event_handler_conext: ScopeEventHandlerContext) -> None:
```

exasol_advanced_analytics_framework/event_handler/context/top_level_event_handler_context.py
```diff
@@ -0,0 +1,219 @@
...
+        )
+        self._child_event_handler_context_list.append(child_event_handler_conext)
+        return child_event_handler_conext
+
+    def _is_child(self, scope_event_handler_context: ScopeEventHandlerContext):
+        result = isinstance(scope_event_handler_context, _ChildEventHandlerContext) and \
+                 scope_event_handler_context._parent == self
+        return result
+
+    def _transfer_object_to(self, object_proxy: ObjectProxy,
```
umitbuyuksahin, line 104: when do you need to transfer object to another context ?

exasol_advanced_analytics_framework/event_handler/context/top_level_event_handler_context.py
```diff
@@ -0,0 +1,219 @@
...
+                            scope_event_handler_conext: ScopeEventHandlerContext):
+        self._check_if_valid()
+        if object_proxy in self._owned_object_proxies:
+            scope_event_handler_conext._own_object(object_proxy)
+            if not self._is_child(scope_event_handler_conext):
+                self._remove_object(object_proxy)
+        else:
+            raise RuntimeError("Object not owned by this ScopeEventHandlerContext.")
+
+    def _remove_object(self, object_proxy: ObjectProxy):
```
umitbuyuksahin, line 114:
```suggestion
    def _remove_object(self, object_proxy: ObjectProxy) -> None:
```

exasol_advanced_analytics_framework/event_handler/context/top_level_event_handler_context.py
```diff
@@ -0,0 +1,219 @@
...
+        self._valid_object_proxies.remove(object_proxy)
+        self._invalid_object_proxies.add(object_proxy)
+        self._parent._release_object(object_proxy)
+
+    def _register_object(self, object_proxy: ObjectProxy):
+        self._check_if_valid()
+        self._valid_object_proxies.add(object_proxy)
+        self._parent._register_object(object_proxy)
+
+    def _is_parent(self, scope_event_handler_context: ScopeEventHandlerContext):
```
umitbuyuksahin, line 203:
```suggestion
    def _is_parent(self, scope_event_handler_context: ScopeEventHandlerContext) -> bool:
```

exasol_advanced_analytics_framework/event_handler/context/top_level_event_handler_context.py
```diff
@@ -0,0 +1,219 @@
...
+        result = self._parent == scope_event_handler_context
+        return result
+
+    def _is_sibling(self, scope_event_handler_context: ScopeEventHandlerContext):
```
umitbuyuksahin, line 207:
```suggestion
    def _is_sibling(self, scope_event_handler_context: ScopeEventHandlerContext) -> bool:
```

exasol_advanced_analytics_framework/event_handler/context/proxy/table_like_proxy.py
```diff
@@ -0,0 +1,14 @@
+from exasol_data_science_utils_python.preprocessing.sql.schema.table_name import TableName
+
+from exasol_advanced_analytics_framework.event_handler.context.proxy.db_object_proxy import DBObjectProxy
+
+
+class TableLikeProxy(DBObjectProxy):
```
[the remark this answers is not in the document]
tkilias: We might want to create other DBObject in the future, e.g. UDF, Connection-Objects, ...

exasol_advanced_analytics_framework/event_handler/context/top_level_event_handler_context.py
tkilias: you are right, probably a leftover

exasol_advanced_analytics_framework/event_handler/context/top_level_event_handler_context.py
tkilias: so you mean we extract it to the parent as is

exasol_advanced_analytics_framework/event_handler/context/top_level_event_handler_context.py
tkilias: so you mean we cal the extracted method from the super class and then we call self._parent_..., yeah, I think, this is a good idea, especially, if it should get more involved in the future.

exasol_advanced_analytics_framework/event_handler/context/top_level_event_handler_context.py
tkilias: In theory, we can build something like a function graph out of event handler, for example, you have a parent function, which first calls child function 1 and then child function 2. In our framework, each function would get a EventHandlerContext. If now assume child function 1 create temporary objects and returns one of them to the parent. Then, we want that all temporary objects except the returned one, are removed after child function 1 ended. This means the return temporary object needs to be transfered from the EventHandlerContext of child function1 which we will release, to another one which won't be released. This means we can either transfer it the the EventHandlerContext of the parent function. If the returned temporary object should get the input for child function 2, we can also transfer it the EventContextHandler of child function 2.

tkilias: I will add a documentation

exasol_advanced_analytics_framework/event_handler/context/top_level_event_handler_context.py
umitbuyuksahin: yes exactly

exasol_advanced_analytics_framework/event_handler/context/top_level_event_handler_context.py
umitbuyuksahin: exactly

exasol_advanced_analytics_framework/event_handler/context/top_level_event_handler_context.py
umitbuyuksahin: thanks

exasol_advanced_analytics_framework/event_handler/context/event_handler_context.py
```diff
@@ -0,0 +1,34 @@
...
+from exasol_advanced_analytics_framework.event_handler.context.proxy.table_proxy import TableProxy
+from exasol_advanced_analytics_framework.event_handler.context.proxy.view_proxy import ViewProxy
+
+
+class EventHandlerContext(ABC):
+    @abc.abstractmethod
+    def get_temporary_table(self) -> TableProxy:
+        """
+        This function registers a new temporary table without creating it.
+        After the release of this context the framworkf will issue a cleanup query.
```
umitbuyuksahin, line 14:
```suggestion
        After the release of this context the framework will issue a cleanup query.
```

exasol_advanced_analytics_framework/event_handler/context/event_handler_context.py
```diff
@@ -0,0 +1,34 @@
...
+        """
+        pass
+
+    @abc.abstractmethod
+    def get_temporary_view(self) -> ViewProxy:
+        """
+        This function registers a new temporary view without creating it.
+        After the release of this context the framworkf will issue a cleanup query.
```
umitbuyuksahin, line 22:
```suggestion
        After the release of this context the framework will issue a cleanup query.
```

exasol_advanced_analytics_framework/event_handler/context/event_handler_context.py
```diff
@@ -0,0 +1,34 @@
...
+        """
+
+        pass
+
+    @abc.abstractmethod
+    def get_temporary_bucketfs_file(self) -> BucketFSLocationProxy:
+        """
+        This function registers a new temporary bucketfs file without creating it.
+        After the release of this context the framworkf will remove it.
```
umitbuyuksahin, line 31:
```suggestion
        After the release of this context the framework will remove it.
```

exasol_advanced_analytics_framework/event_handler/context/scope_event_handler_context.py
```diff
@@ -0,0 +1,49 @@
...
+
+    @abstractmethod
+    def _invalidate(self):
+        pass
+
+    @abstractmethod
+    def transfer_object_to(self, object_proxy: ObjectProxy,
+                           scope_event_handler_context: "ScopeEventHandlerContext"):
+        """
+        This function transfers the ownershio if the object to a different context.
```
umitbuyuksahin, line 39:
```suggestion
        This function transfers the ownership of the object to a different context.
```

exasol_advanced_analytics_framework/event_handler/context/scope_event_handler_context.py
```diff
@@ -0,0 +1,49 @@
...
+        That means, that the object isn't released if this context is released,
+        instread it will be released if the ohter context is released.
```
umitbuyuksahin, line 41:
```suggestion
        instead it will be released if the other transferred context is released.
```

exasol_advanced_analytics_framework/event_handler/context/scope_event_handler_context.py
```diff
@@ -0,0 +1,49 @@
...
+        However, object can be only transferred to the parent, child or sibling context.
```
umitbuyuksahin, line 42:
```suggestion
        However, the object can be only transferred to the parent, child or sibling context.
```

exasol_advanced_analytics_framework/event_handler/context/scope_event_handler_context.py
```diff
@@ -0,0 +1,49 @@
...
+        The first owner is always the context where one of the get_*_object function was called.
+        Transferring the object from one context to another can be used in conjunction with
+        nested event handlers where you want to cleanup after one event handler finished,
+        but want to exchange some temporary objects between the event handlers. For the transfer,
+        is always the parent event handler responsible
```
umitbuyuksahin, line 47:
```suggestion
        but want to exchange some temporary objects between the event handlers. The parent event 
        handler is always responsible for the transfer.
```

umitbuyuksahin (commented):
typos

umitbuyuksahin (approved):

PR #56, opened 2022-08-29
#55 Added Mock Query Handler Runner

### LINKED ISSUE #55: Add MockQueryHandlerRunner
# Background

- To test SQL-based query handlers we need to actually run the query handlers against a database, however using the Lua-based QueryHandlerRunner requires the build of the language container
- To avoid this, we need a Python-based runner which can send the queries to the database via a SQL-Client
- But furthermore, by abstracting the SQL-Client, we can also mock this and compare in unit test the generated queries with our expectation.

tkilias:
Fixes #55


exasol_advanced_analytics_framework/testing/mock_query_handler_runner.py
```diff
@@ -0,0 +1,102 @@
...
+                self.handle_finish()
+                return result.result
+            else:
+                while isinstance(result, Continue):
+                    result = self.handle_continue(result)
+                if isinstance(result, Finish):
+                    self.handle_finish()
+                    return result.result
+                else:
+                    raise RuntimeError("Unknown Result")
```
tomuben, line 48: Does it really need the outer if/else branch?
I mean:

```suggestion
            while isinstance(result, Continue):
                result = self.handle_continue(result)
            if isinstance(result, Finish):
                self.handle_finish()
                return result.result
            else:
                raise RuntimeError("Unknown Result")
```

exasol_advanced_analytics_framework/testing/mock_query_handler_runner.py
```diff
@@ -0,0 +1,102 @@
...
+            raise RuntimeError(f"Execution of query handler {self._state.query_handler} failed.") from e
+
+    def handle_continue(self, result: Continue):
+        self.release_and_create_new_return_query_query_handler_context()
+        self.cleanup_query_handler_context()
+        for query in result.query_list:
+            self._sql_executor.execute(query.query_string)
+        input_query_result = self.run_input_query(result)
+        result = self._state.query_handler.handle_query_result(input_query_result)
+        return result
```
tomuben, line 60:
```suggestion
        return self._state.query_handler.handle_query_result(input_query_result)
```

exasol_advanced_analytics_framework/testing/mock_query_handler_runner.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def run_input_query(self, result: Continue):
```
tomuben, line 62: add an type hint for the return type

exasol_advanced_analytics_framework/testing/mock_query_handler_runner.py
```diff
@@ -0,0 +1,102 @@
...
+                if isinstance(result, Finish):
+                    self.handle_finish()
+                    return result.result
+                else:
+                    raise RuntimeError("Unknown Result")
+        except Exception as e:
+            self.handle_finish()
+            raise RuntimeError(f"Execution of query handler {self._state.query_handler} failed.") from e
+
+    def handle_continue(self, result: Continue):
```
tomuben, line 53: add type hint for return type.

exasol_advanced_analytics_framework/testing/mock_query_handler_runner.py
```diff
@@ -0,0 +1,102 @@
...
+        result = self._state.query_handler.handle_query_result(input_query_result)
+        return result
+
+    def run_input_query(self, result: Continue):
+        input_query_view, input_query = self._wrap_return_query(result.input_query)
+        self._sql_executor.execute(input_query_view)
+        input_query_result_set = self._sql_executor.execute(input_query)
+        if input_query_result_set.columns() != result.input_query.output_columns:
+            raise RuntimeError(f"Specified columns {result.input_query.output_columns} of the input query "
+                               f"are equal to the actual received columns {input_query_result_set.columns()}")
```
tomuben, line 68:
```suggestion
            raise RuntimeError(f"Specified columns {result.input_query.output_columns} of the input query "
                               f"are not equal to the actual received columns {input_query_result_set.columns()}")
```

exasol_advanced_analytics_framework/testing/mock_query_handler_runner.py
```diff
@@ -0,0 +1,102 @@
...
+        if self._state.input_query_query_handler_context is not None:
+            self._state.input_query_query_handler_context.release()
+        self._state.top_level_query_handler_context.release()
+        self.cleanup_query_handler_context()
+
+    def cleanup_query_handler_context(self):
+        cleanup_query_list = \
+            self._state.top_level_query_handler_context.cleanup_released_object_proxies()
+        for query in cleanup_query_list:
+            self._sql_executor.execute(query.query_string)
```
tomuben, line 84: This code snippet appears several times in this class, create helper function:
def execute_query_list(query_list: Iterator[str]) -> None:
       for query in query_list:
            self._sql_executor.execute(query.query_string)

exasol_advanced_analytics_framework/testing/mock_query_handler_runner.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def release_and_create_new_return_query_query_handler_context(self):
```
tomuben, line 86: is the double "query" in the name by intention?

```suggestion
    def release_and_create_new_return_query_handler_context(self):
```

tomuben: Also add a type hint for the return type

tests/unit_tests/query_handler_runner/test_mock_query_handler_runner.py
```diff
@@ -0,0 +1,467 @@
...
+from exasol_advanced_analytics_framework.query_handler.query_handler import QueryHandler
+from exasol_advanced_analytics_framework.query_handler.result import Continue, Finish
+from exasol_advanced_analytics_framework.query_result.query_result import QueryResult
+from exasol_advanced_analytics_framework.testing.mock_query_handler_runner import MockQueryHandlerRunner
+
+
+@pytest.fixture()
+def temporary_schema_name():
+    temporary_schema_name = "temp_schema_name"
+    return temporary_schema_name
```
tomuben, line 27:
```suggestion
    return "temp_schema_name"
```

tests/unit_tests/query_handler_runner/test_mock_query_handler_runner.py
```diff
@@ -0,0 +1,467 @@
...
+
+    def start(self) -> Union[Continue, Finish[TestOutput]]:
+        return Finish[TestOutput](TestOutput(self._parameter))
+
+    def handle_query_result(self, query_result: QueryResult) -> Union[Continue, Finish[TestOutput]]:
+        pass
+
+
+def test_start_finish(top_level_query_handler_context):
+    sql_executor = MockSQLExecutor(result_sets=[])
```
tomuben, line 62: Would be nice to have a short description about what the test actually does.

tests/unit_tests/query_handler_runner/test_mock_query_handler_runner.py
```diff
@@ -0,0 +1,467 @@
...
+    def start(self) -> Union[Continue, Finish[TestOutput]]:
+        self._query_handler_context.get_temporary_table()
+        return Finish[TestOutput](TestOutput(self._parameter))
+
+    def handle_query_result(self, query_result: QueryResult) -> Union[Continue, Finish[TestOutput]]:
+        pass
+
+
+def test_start_finish_cleanup_queries(temporary_schema_name, top_level_query_handler_context):
+    sql_executor = MockSQLExecutor(result_sets=[MockResultSet()])
```
tomuben, line 88: Would be nice to have a short description about what the test actually does.

tomuben (changes requested):
check my comments

exasol_advanced_analytics_framework/testing/mock_query_handler_runner.py
tkilias: Nie, that you wrote, you are probably right. This a leftover from the construction. I start often case by case

exasol_advanced_analytics_framework/testing/mock_query_handler_runner.py
tkilias: I prefer to assign the result of a function first to variable, because this make debugging easier

exasol_advanced_analytics_framework/testing/mock_query_handler_runner.py
tkilias: I would prefer not to do this, because query list has a dedicated meaning in this algorithm. We can rename the method to execute_queries or so, then it should be possible

exasol_advanced_analytics_framework/testing/mock_query_handler_runner.py
tkilias: Regarding the name, yes it is actually intended. We have something that is called return_query or input_query (need to change at least this name) and a other thing which is called query_handler_context. I think, the best would be if we rename it to the query_handler_context_of_input_query

exasol_advanced_analytics_framework/testing/mock_query_handler_runner.py
tkilias: done

exasol_advanced_analytics_framework/testing/mock_query_handler_runner.py
tkilias: done

exasol_advanced_analytics_framework/testing/mock_query_handler_runner.py
tkilias: done

exasol_advanced_analytics_framework/testing/mock_query_handler_runner.py
tkilias: done

tests/unit_tests/query_handler_runner/test_mock_query_handler_runner.py
tkilias: done

tests/unit_tests/query_handler_runner/test_mock_query_handler_runner.py
tkilias: done

exasol_advanced_analytics_framework/testing/mock_query_handler_runner.py
```diff
@@ -80,10 +76,13 @@ def handle_finish(self):
...
             self._state.top_level_query_handler_context.cleanup_released_object_proxies()
-        for query in cleanup_query_list:
+        self.execute_query(cleanup_query_list)
+
+    def execute_query(self, queries: List[Query]):
+        for query in queries:
             self._sql_executor.execute(query.query_string)
 
-    def release_and_create_new_return_query_query_handler_context(self):
+    def release_and_create_query_handler_context_of_input_query(self):
```
tomuben, line 85: the name is much better

tests/unit_tests/query_handler_runner/test_mock_query_handler_runner.py
```diff
@@ -145,11 +161,18 @@ def handle_query_result(self, query_result: QueryResult) -> Union[Continue, Fini
 
 
 def test_continue_finish(temporary_schema_name, top_level_query_handler_context):
+    """
+    This tests runs a query handler which returns Continue result from the start method
+    and expect to be handle_query_result to be called. Further, it expects that
+    handle_query_result can access the columns in the resultset which where defined
+    in the input_query.
```
tomuben, line 168:
```suggestion
    This tests runs a query handler which returns Continue result from the start method
    and expect handle_query_result to be called. Further, it expects that
    handle_query_result can access the columns in the resultset which where defined
    in the input_query.
```

tests/unit_tests/query_handler_runner/test_mock_query_handler_runner.py
```diff
@@ -424,6 +469,13 @@ def handle_query_result(self, query_result: QueryResult) -> Union[Continue, Fini
 
 
 def test_continue_cleanup_continue_finish(temporary_schema_name, top_level_query_handler_context):
+    """
+    This tests runs a query handler which creates the temporary table of a child query context manager.
+    Then it returns a Continue results, such that handle_query_result will be called. During the call to
+    handle_query_result we release child query context manager and returns a Continue results such that
+    handle_query_result get called again to return a Finish result.
+    We expect that the cleanup of the temporary happens between the first and second call to handle_query_result.
```
tomuben, line 477:
```suggestion
    This tests runs a query handler which creates the temporary table of a child query context manager.
    Then it returns a Continue result, such that handle_query_result will be called. During the call to
    handle_query_result we release child query context manager and return a Continue result such that
    handle_query_result gets called again, which then returns a Finish result.
    We expect that the cleanup of the temporary happens between the first and second call to handle_query_result.
```

tests/unit_tests/query_handler_runner/test_mock_query_handler_runner.py
```diff
@@ -323,44 +358,54 @@ def test_continue_error_cleanup_queries(temporary_schema_name, top_level_query_h
...
             return Continue(query_list=[], input_query=input_query)
         else:
             return Finish[TestOutput](TestOutput(self._parameter))
 
 
 def test_continue_continue_finish(temporary_schema_name, top_level_query_handler_context):
+    """
+    This tests runs a query handler which returns Continue from the first call to handle_query_result method
+    and the second time it returns Finish. We expect two input queries to be executed one per Continue and
+    in the end the result should be returned.
```
tomuben, line 396:
```suggestion
    This tests runs a query handler which returns Continue from the first call to handle_query_result method
    and the second time it returns Finish. We expect two input queries to be executed; one per Continue and
    in the end the result should be returned.
```

tomuben (commented):
check my comments

tomuben (approved):

PR #57, opened 2022-08-30
#53: Fix that the QueryHandlerRunnerUDF doesn't serialize or deserialize

### LINKED ISSUE #53: QueryHandlerRunnerUDF doesn't serialize the final_result to json
# Background

- The ResultType of the UDFQueryHandler assumes a Dict[str,Any] as return value and it expects that this will be serialized to json from the QueryHandlerRunnerUDF
- However, currently, we transform the result into a string and return this instead

tkilias:
Fixes #53


exasol_advanced_analytics_framework/udf_framework/udf_query_handler.py
```diff
@@ -1,7 +1,10 @@
 from typing import Dict, Any
 
+from exasol_advanced_analytics_framework.query_handler.context.scope_query_handler_context import \
+    ScopeQueryHandlerContext
 from exasol_advanced_analytics_framework.query_handler.query_handler import QueryHandler
 
 
-class UDFQueryHandler(QueryHandler[Dict[str, Any], Dict[str, Any]]):
+class UDFQueryHandler(QueryHandler[str, str]):
```
tomuben, line 8: Isn't that an abstract class, too?

tomuben: If so, inherit from `ABC`

exasol_advanced_analytics_framework/udf_framework/udf_query_handler_factory.py
```diff
@@ -0,0 +1,12 @@
...
+from exasol_advanced_analytics_framework.query_handler.context.scope_query_handler_context import \
+    ScopeQueryHandlerContext
+from exasol_advanced_analytics_framework.udf_framework.udf_query_handler import UDFQueryHandler
+
+
+class UDFQueryHandlerFactory(ABC):
+
+    @abstractmethod
+    def create(self, parameter: str, query_handler_context: ScopeQueryHandlerContext) -> UDFQueryHandler:
+        raise NotImplementedError()
```
tomuben, line 12: I believe there is no need to raise the exception by hand,  method annoted with @abstractmethod automatically raise this exception
```suggestion
        pass
```

or better

```suggestion
        """ description """
```

exasol_advanced_analytics_framework/udf_framework/udf_query_handler.py
```diff
@@ -1,7 +1,10 @@
...
     pass
```
tomuben, line 9: Learned from Nico that it's better to use a docstring in this case :) 
```suggestion
    """ description """
```

tests/integration_tests/test_query_loop_integration.py
```diff
@@ -47,12 +47,13 @@ def test_query_loop_integration_with_one_iteration(
...
 
-    assert result[0][0] == str(FINAL_RESULT)
+    assert result[0][0] == FINAL_RESULT
 
 
 def test_query_loop_integration_with_one_iteration_with_not_released_child_query_handler_context(
         setup_database, upload_language_container):
-    # start a new db session, to isolate the EXECUTE SCRIPT and the QueryHandler queries into its own session for easer retrieval
+    # start a new db session, to isolate the EXECUTE SCRIPT and the QueryHandler queries
+    # into its own session for easer retrieval
```
tomuben, line 56:
```suggestion
    # into its own session for easier retrieval
```

tests/integration_tests/test_query_loop_integration.py
```diff
@@ -90,7 +91,8 @@ def test_query_loop_integration_with_one_iteration_with_not_released_child_query
 
 def test_query_loop_integration_with_one_iteration_with_not_released_temporary_object(
         setup_database, upload_language_container):
-    # start a new db session, to isolate the EXECUTE SCRIPT and the QueryHandler queries into its own session for easer retrieval
+    # start a new db session, to isolate the EXECUTE SCRIPT and the QueryHandler queries
+    # into its own session for easer retrieval of the audit log
```
tomuben, line 95:
```suggestion
    # into its own session, for easier retrieval of the audit log
```

tests/integration_tests/test_query_loop_integration.py
```diff
@@ -132,12 +134,15 @@ def test_query_loop_integration_with_one_iteration_with_not_released_temporary_o
...
+    for query in executed_queries:
+        print("executed_query: ", query)
     assert table_cleanup_query
 
 
 def test_query_loop_integration_with_two_iteration(
         setup_database, upload_language_container):
-    # start a new db session, to isolate the EXECUTE SCRIPT and the QueryHandler queries into its own session for easer retrieval
+    # start a new db session, to isolate the EXECUTE SCRIPT and the QueryHandler queries
+    # into its own session for easer retrieval of the audit log
```
tomuben, line 145:
```suggestion
    # into its own session, for easier retrieval of the audit log
```

tests/test_package/test_query_handlers/query_handler_test.py
```diff
@@ -52,34 +60,51 @@ def start(self) -> Union[Continue, Finish[Dict[str, Any]]]:
...
         return Finish(result=FINAL_RESULT)
 
-    def handle_query_result(self, query_result: QueryResult) -> Union[Continue, Finish[Dict[str, Any]]]:
+    def handle_query_result(self, query_result: QueryResult) -> Union[Continue, Finish[str]]:
         pass
+
+class QueryHandlerWithOneIterationWithNotReleasedTemporaryObjectFactory(UDFQueryHandlerFactory):
+
+    def create(self, parameter: str, query_handler_context: ScopeQueryHandlerContext) -> UDFQueryHandler:
+        return QueryHandlerWithOneIterationWithNotReleasedTemporaryObject(parameter, query_handler_context)
```
tomuben, line 110: the factories are all the same, only difference is the returned type. looks like a lot of boiler plate code.

```suggestion
class QueryHandlerTestFactory(UDFQueryHandlerFactory):
    def __init__(self, returned_type):
        self._returned_type = returned_type
        
    def create(self, parameter: str, query_handler_context: ScopeQueryHandlerContext) -> UDFQueryHandler:
        return self._returned_type(parameter, query_handler_context)
 
 QueryHandlerWithOneIterationWithNotReleasedTemporaryObjectFactory = QueryHandlerTestFactory(QueryHandlerWithOneIterationWithNotReleasedTemporaryObject)
 QueryHandlerWithOneIterationWithNotReleasedChildQueryHandlerContextFactory = QueryHandlerTestFactory(QueryHandlerWithOneIterationWithNotReleasedChildQueryHandlerContext)
 ...
```

tests/unit_tests/udf_framework/mock_query_handlers.py
```diff
@@ -57,49 +64,72 @@ def start(self) -> Union[Continue, Finish[Dict[str, Any]]]:
...
 
-    def handle_query_result(self, query_result: QueryResult) -> Union[Continue, Finish[Dict[str, Any]]]:
+    def handle_query_result(self, query_result: QueryResult) -> Union[Continue, Finish[str]]:
         pass
 
 
+class QueryHandlerTestWithOneIterationAndTempTableFactory(UDFQueryHandlerFactory):
+
+    def create(self, parameter: str, query_handler_context: ScopeQueryHandlerContext) -> UDFQueryHandler:
+        return QueryHandlerTestWithOneIterationAndTempTable(parameter, query_handler_context)
```
tomuben, line 93: same as above

tomuben (changes requested):
check my comments

tests/test_package/test_query_handlers/query_handler_test.py
tkilias: The reason is, we inject the factories by name and not by value or reference. They will be created in the UDF from their name.

tkilias: Before we injected the Query handler directly by the class name without the factories, however this didn't allow us to wrap the actual query handler

exasol_advanced_analytics_framework/udf_framework/udf_query_handler.py
tkilias: QueryHandler is an abstract class and pass on the abstract status to subclasses https://stackoverflow.com/questions/48814211/python3-how-does-one-define-an-abstract-subclass-from-an-existing-abstract-cla

tomuben (approved):

PR #63, opened 2022-09-02
#62: Fix bug in ScopeQueryHandlerContext transfer_object_to

### LINKED ISSUE #62: Potential bug in _ScopeQueryHandlerContextBase._transfer_object_to
# Background

- At the following code position, it looks like, we only remove the object from the current context, if the other context is not a child (means a sibling.). However, we need to remove the object regardless of the relationship from the owner list of the current context,
https://github.com/exasol/advanced-analytics-framework/blob/094707933b3f0af5f81812154c3897682d4ee166/exasol_advanced_analytics_framework/query_handler/context/top_level_query_handler_context.py#L142

tkilias:
Fixes #62

### DIFF
--- a/doc/changes/changes_0.1.0.md
+++ b/doc/changes/changes_0.1.0.md
@@ -21,6 +21,7 @@
 ### Bug Fixes
 
   - #8: Renamed master branch to main
+  - #62: Fix bug in ScopeQueryHandlerContext transfer_object_to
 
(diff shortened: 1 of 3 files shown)


tomuben (approved):

tomuben (changes requested):
update changelog

tomuben (approved):

PR #65, opened 2022-09-05
#64: Remove _check_if_valid calls in __eq__ and __hash__ for DBObjectNameProxy

### LINKED ISSUE #64: Fix __hash__ DBObjectNameProxy
# Background

- This is, currently, the hash function of DBObjectNameProxy
```
    def __hash__(self):
        self._check_if_valid()
        return hash(id(self))

```

- It has two issues
- First we use self._check_if_valid() which fails when the DBObject was release. If we store DBObject in Dicts as key or in Sets, this can cause that we can't remove these objects anymore from the container.
  - The same issues has __eq__
-  The second, issue is that it uses the id which might change after pickle unpickle which is a common case for us.

tkilias:
Fixes #64

### DIFF
--- a/doc/changes/changes_0.1.0.md
+++ b/doc/changes/changes_0.1.0.md
@@ -22,7 +22,7 @@
 
   - #8: Renamed master branch to main
   - #62: Fix bug in ScopeQueryHandlerContext transfer_object_to
-
+  - #64: Remove _check_if_valid calls in __eq__ and __hash__ for DBObjectNameProxy
 ### Refactoring
 
  - #42: Updated dependencies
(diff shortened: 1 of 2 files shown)


doc/changes/changes_0.1.0.md
```diff
@@ -22,7 +22,7 @@
 
   - #8: Renamed master branch to main
   - #62: Fix bug in ScopeQueryHandlerContext transfer_object_to
-
+  - #64: Remove _check_if_valid calls in __eq__ and __hash__ for DBObjectNameProxy
```
tomuben, line 25: __eq__ and __hash__ are interpreted as markdown formattings....

tomuben (dismissed):

tomuben (approved):

Issue #68, opened 2022-10-05
Methods called in BucketFSLocationProxy.cleanup can fail and stop the cleanup


tkilias:
# Background:

```
    def cleanup(self):
        if self._is_valid:
            raise Exception("Cleanup of BucketFSLocationProxy only allowed after release.")
        for file in self._bucketfs_location.list_files_in_bucketfs(""):
            self._bucketfs_location.delete_file_in_bucketfs(file)
```
- list_files_in_bucketfs can fail, if the location wasn't used for uploading
- delete_file_in_bucketfs can fail, but it shouldn't prevent use from deleting the other files

https://github.com/exasol/advanced-analytics-framework/blob/6f92ae59456560718f5e4ad608e76c8f31b1fb57/exasol_advanced_analytics_framework/query_handler/context/proxy/bucketfs_location_proxy.py#L19

PR #74, opened 2022-10-06
#68: Methods called in BucketFSLocationProxy.cleanup can fail and stop the cleanup

tkilias:
Fixes #74

### DIFF
--- a/exasol_advanced_analytics_framework/query_handler/context/proxy/bucketfs_location_proxy.py
+++ b/exasol_advanced_analytics_framework/query_handler/context/proxy/bucketfs_location_proxy.py
@@ -34,6 +34,8 @@ class BucketFSLocationProxy(ObjectProxy):
         files = []
         try:
             files = self._bucketfs_location.list_files_in_bucketfs("")
+        except FileNotFoundError as e:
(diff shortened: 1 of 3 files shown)


exasol_advanced_analytics_framework/query_handler/context/proxy/bucketfs_location_proxy.py
```diff
@@ -16,5 +20,20 @@ def bucketfs_location(self) -> AbstractBucketFSLocation:
...
+            self._bucketfs_location.delete_file_in_bucketfs(file)
+        except Exception as e:
+            LOGGER.error(f"Failed to remove {file}, got exception", exc_info=True)
+
+    def _list_files(self):
+        files = []
+        try:
+            files = self._bucketfs_location.list_files_in_bucketfs("")
+        except Exception as e:
+            LOGGER.debug(f"Got exception during listing files in temporary BucketFSLocation", exc_info=True)
```
tomuben, line 38: why only debug? Shouldn't be at least a warning?

tkilias: ah,  I catch Exception, I wanted to catch FileNotFound and make a debug and for Exception an error

tomuben (dismissed):

tomuben (approved):

PR #77, opened 2022-10-10
#76: Added get_connection to query_handler_context

### LINKED ISSUE #76: Add get_connection to query_handler_context
# Background

- Currently, query_handler can't access the connection object to retrieve credentials.
- UDFs provide the get_connection function for this, we need to forward this to the query_handlers
- Because, we already inject the query_handler_context and it is more or less for responsible for the communication with the outside, it is probably the best point to add this functionality.

tkilias:
Fixes #76


exasol_advanced_analytics_framework/query_handler/context/scope_query_handler_context.py
```diff
@@ -1,9 +1,33 @@
...
 from exasol_advanced_analytics_framework.query_handler.context.query_handler_context import QueryHandlerContext
 from exasol_advanced_analytics_framework.query_handler.context.proxy.object_proxy import ObjectProxy
 
 
+class Connection(ABC):
+
+    @property
+    @abstractmethod
+    def name(self) -> str:
+        pass
```
Nicoretti, line 13: If there is need to communicate additional information about the interface, I personally prefer the doc string based empty definition.

```suggestion
    @property
    @abstractmethod
    def name(self) -> str:
        """ <some comment/docstring> ""
```

pyproject.toml
```diff
@@ -25,6 +25,7 @@ exasol-data-science-utils-python = {git = "https://github.com/exasol/data-scienc
 click = "^8.0.4"
 pyexasol = "^0.25.0"
 pandas = "^1.1.0"
+password = "^0.2"
```
Nicoretti, line 28: What exactly do yo need this for?

Nicoretti (approved):

pyproject.toml
tkilias: good catch, I think, that was something the IDE did by accident

tkilias: will remove

Nicoretti (approved):

PR #78, opened 2022-10-17
#79: Add UDF local discovery

### LINKED ISSUE #79: Add UDF Local Discovery
Background:

Some algorithms need communication between UDF Instances of a UDF Call for example ML algorithms or some HPC algorithms. To achieve this, we need first to discovery all UDF instances of a UDF call. We can have many instances in a cluster, such that we need to do first a local discovery per node.

tkilias:
Fixes #79


doc/design/udf_discovery_and_communication.rst
```diff
@@ -0,0 +1,63 @@
...
+  - Send a ReadyToReceiveMessage with our connection information for the receiving socket
+    over our sending socket for it, in case we didn't get its PingMessage.
+
+.. image:: peer_communicator_state_machine.drawio.png
+
+********
+Details:
+********
+
+We seperated the Local Discovery into two components. The component LocalDiscovery implements
```
tomuben, line 39: You use the term Local Discovery for the whole communication process, but also for one component. This could be a bit confusing.

tomuben:
Maybe it would be good to add a general overview of the discovery, something like:
udf-discovery drawio

exasol_advanced_analytics_framework/udf_communication/background_listener.py
```diff
@@ -0,0 +1,110 @@
...
+from exasol_advanced_analytics_framework.udf_communication.ip_address import IPAddress
+from exasol_advanced_analytics_framework.udf_communication.messages import Message, StopMessage, RegisterPeerMessage, \
+    MyConnectionInfoMessage
+from exasol_advanced_analytics_framework.udf_communication.peer import Peer
+from exasol_advanced_analytics_framework.udf_communication.serialization import deserialize_message, serialize_message
+
+LOGGER: FilteringBoundLogger = structlog.get_logger()
+
+
+class BackgroundListener:
```
tomuben, line 20: Actually, I am thinking, maybe it would be good to rename that class to something more specific, maybe "PeerCommunicationBackgroundListener".

exasol_advanced_analytics_framework/udf_communication/background_peer_state.py
```diff
@@ -0,0 +1,94 @@
...
+        self._context = context
+        self._peer_can_receive_from_us = False
+        self._last_send_ready_to_receive_timestamp_in_seconds: Optional[float] = None
+        self._create_receive_socket()
+        self._send_we_are_ready_to_receive()
+
+    def _create_receive_socket(self):
+        self._receive_socket = self._context.socket(zmq.PAIR)
+        receive_socket_address = get_peer_receive_socket_name(self._peer)
+        self._receive_socket.bind(receive_socket_address)
```
tomuben, line 38: that name is actually a bit confusing. As far as I understood that is the data socket between frontend and backend for a specific peer. Maybe rename it to "frontend_data_socket"

exasol_advanced_analytics_framework/udf_communication/frontend_peer_state.py
```diff
@@ -0,0 +1,64 @@
...
+                 context: zmq.Context,
+                 peer: Peer):
+        self._my_connection_info = my_connection_info
+        self._peer = peer
+        self._context = context
+        self._peer_is_ready = False
+        self._create_receive_socket()
+
+    def _create_receive_socket(self):
+        self._receive_socket = self._context.socket(zmq.PAIR)
```
tomuben, line 29: same here, the name is a bit confusing. maybe rename to "backend_data_socket"

exasol_advanced_analytics_framework/udf_communication/local_discovery.py
```diff
@@ -0,0 +1,56 @@
...
+        discovery_timeout_in_ns = self._discovery_timeout_in_minutes * 10 ** 9
+        result = time_difference_ns > discovery_timeout_in_ns
+        return result
+
+    def _discover_peers(self):
+        self._send_ping()
+        begin_time_ns = time.monotonic_ns()
+        while not self._should_discovery_end(begin_time_ns):
+            self._receive_ping()
+            self._send_ping()
```
tomuben, line 35: This way you process only one ping message and send a ping. Wouldn't it be better to process all received ping messages in zmq queue before sending the ping?

tkilias:
> Maybe it would be good to add a general overview of the discovery, something like: udf-discovery drawio

Yeah thought about it, but didn't yet have the right idea how to visualize it, Maybe your sketch gives me an idea.

doc/design/udf_discovery_and_communication.rst
tkilias: Maybe we split into Discovery and Cluster Formation. Or rename the component into LocalDiscoveryStrategy. However, in general, the handshakes the PeerCommunicator does, are part of the Discovery, because only afterwards, it is guaranteed that all peers know each other. Need to think about it.

exasol_advanced_analytics_framework/udf_communication/local_discovery.py
```diff
@@ -0,0 +1,56 @@
...
+class LocalDiscovery:
+
+    def __init__(self,
+                 discovery_timeout_in_seconds: int,
+                 peer_communicator: PeerCommunicator,
+                 local_discovery_socket: LocalDiscoverySocket
+                 ):
+        self._local_discovery_socket = local_discovery_socket
+        self._peer_communicator = peer_communicator
+        self._discovery_timeout_in_minutes = discovery_timeout_in_seconds
```
tomuben, line 20: look wrong. shouldn't be:
```suggestion
        self._discovery_timeout_in_minutes = discovery_timeout_in_seconds / 60
```
?

exasol_advanced_analytics_framework/udf_communication/local_discovery.py
```diff
@@ -0,0 +1,56 @@
...
+        self._discover_peers()
+
+    def _is_timeout(self, begin_time_ns: int):
+        current_time_ns = time.monotonic_ns()
+        time_difference_ns = current_time_ns - begin_time_ns
+        discovery_timeout_in_ns = self._discovery_timeout_in_minutes * 10 ** 9
```
tomuben, line 26: if you only use the timeout in ns, why do you save the timeout in the member variable in minutes?

tests/udf_communication/peer_communication/test_add_peer.py
```diff
@@ -0,0 +1,86 @@
...
+    except Exception as e:
+        traceback.print_exc()
+        logger.exception("Exception during test", exception=e)
+
+
+def test():
+    group = f"{time.monotonic_ns()}"
+    logger = LOGGER.bind(group=group, location="test")
+    logger.info("start")
+    number_of_instances = 10
```
tomuben, line 60: Does it make sense to use parametrized test here, maybe with 1, 10 and 100 instances?

tests/udf_communication/peer_communication/test_send_recv.py
```diff
@@ -0,0 +1,77 @@
...
+    com = PeerCommunicator(
+        name=name,
+        number_of_peers=number_of_instances,
+        listen_ip=listen_ip,
+        group_identifier=group_identifier)
+    queue.put(com.my_connection_info)
+    peer_connection_infos = queue.get()
+    for index, connection_infos in peer_connection_infos.items():
+        com.register_peer(connection_infos)
+        time.sleep(random.random() / 10)
```
tomuben, line 39: add comment that this make the process sleep a random amount of time between 0 and 100ms

exasol_advanced_analytics_framework/udf_communication/local_discovery.py
tkilias: should be fixed with rework of the LocalDiscovery

exasol_advanced_analytics_framework/udf_communication/local_discovery.py
tkilias: should be fixed with rework of the LocalDiscovery

exasol_advanced_analytics_framework/udf_communication/local_discovery.py
tkilias: should be fixed with rework of the LocalDiscovery

exasol_advanced_analytics_framework/udf_communication/background_peer_state.py
tkilias: I now what you mean, however I choose this name, because the frontend will only use this socket to receive message from a specific peer. Here, we will send to the socket while forwarding the payload messages to the frontend.

exasol_advanced_analytics_framework/udf_communication/frontend_peer_state.py
tkilias: I would argue that receive socket is here actually better, because its only purpose is that we receive over it payload messages from the specific peer, that the backend forwards them to us is a detail

doc/design/udf_discovery_and_communication.rst
```diff
@@ -0,0 +1,73 @@
...
+*********
+Overview:
+*********
+
+- UDP Broadcast for the initial PingMessage with connection information for the receiving socket of the reliable network
+- After receiving PingMessage:
+
+  - Create sending socket for peer to its receiving socket of the reliable network
+  - Send a ReadyToReceiveMessage with our connection information for the receiving socket
+    over our sending socket for it to inform the peer that we are ready to receive from it
```
tomuben, line 17:
```suggestion
  - Send a ReadyToReceiveMessage with our connection information including the receiving socket port
    over our sending socket for it to inform the peer that we are ready to receive from it
```

doc/design/udf_discovery_and_communication.rst
```diff
@@ -0,0 +1,73 @@
...
+  - Send a ReadyToReceiveMessage with our connection information for the receiving socket
+    over our sending socket for it, in case we didn't get its PingMessage.
+  - If we didn't get a ReadyToReceiveMessage message after a certain time , yet, we send a AreYouReadyToReceiveMessage to a discovered peer
+    - This should prevent a stuck handshake if we should lose a ReadyToReceiveMessage for whatever reason
+
+- After receiving AreYouReadyToReceiveMessage:
+
+  - If not yet discovered, we create a sending socket for the peer to its receiving socket of the reliable network
+  - Send a ReadyToReceiveMessage with our connection information for the receiving socket
+    over our sending socket for it, in case we didn't get its PingMessage.
```
tomuben, line 31:
```suggestion
  - Send a ReadyToReceiveMessage with our connection information including the receiving socket port
    over our sending socket for it, in case we didn't get its PingMessage.
```

exasol_advanced_analytics_framework/udf_communication/local_discovery_socket.py
```diff
@@ -0,0 +1,39 @@
...
+        self._udp_socket = socket.socket(socket.AF_INET, socket.SOCK_DGRAM, socket.IPPROTO_UDP)
+        self._udp_socket.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
+        self._udp_socket.setsockopt(socket.SOL_SOCKET, socket.SO_BROADCAST, 1)
+        self._udp_socket.bind((self._broadcast_ip.ip_address, self._port.port))
+
+    def send(self, message: bytes):
+        self._udp_socket.sendto(message, (self._broadcast_ip.ip_address, self._port.port))
+
+    def recvfrom(self, timeout_in_seconds: float) -> bytes:
+        if timeout_in_seconds<0.0:
```
tomuben, line 22:
```suggestion
        if timeout_in_seconds < 0.0:
```

tomuben: At least error message suggests that timeout_in_seconds must not be equal to 0.0

exasol_advanced_analytics_framework/udf_communication/local_discovery.py
```diff
@@ -1,52 +1,87 @@
...
+        result = self._peer_communicator.are_all_peers_connected() or self._has_discovery_timed_out(begin_time_ns)
         return result
 
-    def _receive_ping(self):
+    def _receive_pings(self, begin_time_ns: int):
+        timeout_in_seconds = self._compute_receive_timeout_in_seconds(begin_time_ns)
+        while True:
+            serialized_message = self._receive_message(timeout_in_seconds)
+            if serialized_message is not None:
+                timeout_in_seconds = self.handle_serialized_message(serialized_message)
```
tomuben, line 58: var timeout_in_seconds is not used

tests/udf_communication/peer_communication/test_send_recv.py
tkilias: removed it, makes the test to slow, this likely get tested anyway by the test_local_discovery.py. Else, we could also add a specific test for this.

tests/udf_communication/peer_communication/test_add_peer.py
tkilias: see my last changes

exasol_advanced_analytics_framework/udf_communication/background_listener.py
tkilias: moved PeerCommunicator and BackgroundListener to the peer_communicator module, because the names got otherwise to long

exasol_advanced_analytics_framework/udf_communication/background_peer_state.py
tomuben: Ok, that's fine.

exasol_advanced_analytics_framework/udf_communication/frontend_peer_state.py
tomuben: Ok, that's fine.

tomuben (dismissed):

tomuben (approved):

PR #89, opened 2022-12-05
#88: Introduce an abstraction for ZMQ in UDF Communication

### LINKED ISSUE #88: Introduce an abstraction for ZMQ in UDF Communication
#  Background

- Currently, the implementation directly uses ZMQ. 
- However, this complicates unit tests.
- To simplify unit tests for the protocol, we need to replace the zmq usage with an abstraction.

tkilias:
Fixes #88

### DIFF
--- a/exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_listener_interface.py
+++ b/exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_listener_interface.py
@@ -3,7 +3,6 @@ import traceback
 from typing import Optional, Iterator
 
 import structlog
-import zmq
 from structlog.types import FilteringBoundLogger
 
 from exasol_advanced_analytics_framework.udf_communication.connection_info import ConnectionInfo
@@ -14,6 +13,8 @@ from exasol_advanced_analytics_framework.udf_communication.peer import Peer
 from exasol_advanced_analytics_framework.udf_communication.peer_communicator.background_listener_thread import \
     BackgroundListenerThread
 from exasol_advanced_analytics_framework.udf_communication.serialization import deserialize_message, serialize_message
+from exasol_advanced_analytics_framework.udf_communication.socket_factory.abstract_socket_factory import SocketFactory, \
+    SocketType, Socket, PollerFlag
 
 LOGGER: FilteringBoundLogger = structlog.get_logger()
 
@@ -22,7 +23,7 @@ class BackgroundListenerInterface:
 
     def __init__(self,
                  name: str,
-                 context: zmq.Context,
+                 socket_factory: SocketFactory,
                  listen_ip: IPAddress,
                  group_identifier: str):
         self._name = name
@@ -32,12 +33,12 @@ class BackgroundListenerInterface:
             name=self._name,
             group_identifier=group_identifier
         )
-        out_control_socket_address = self._create_out_control_socket(context)
-        in_control_socket_address = self._create_in_control_socket(context)
+        out_control_socket_address = self._create_out_control_socket(socket_factory)
+        in_control_socket_address = self._create_in_control_socket(socket_factory)
         self._my_connection_info: Optional[ConnectionInfo] = None
         self._background_listener_run = BackgroundListenerThread(
             name=self._name,
-            context=context,
+            socket_factory=socket_factory,
             listen_ip=listen_ip,
             group_identifier=group_identifier,
             out_control_socket_address=out_control_socket_address,
@@ -48,14 +49,14 @@ class BackgroundListenerInterface:
         self._thread.start()
         self._set_my_connection_info()
 
-    def _create_in_control_socket(self, context) -> str:
-        self._in_control_socket: zmq.Socket = context.socket(zmq.PAIR)
+    def _create_in_control_socket(self, socket_factory: SocketFactory) -> str:
+        self._in_control_socket: Socket = socket_factory.create_socket(SocketType.PAIR)
         in_control_socket_address = f"inproc://BackgroundListener_in_control_socket{id(self)}"
         self._in_control_socket.bind(in_control_socket_address)
         return in_control_socket_address
 
-    def _create_out_control_socket(self, context) -> str:
-        self._out_control_socket: zmq.Socket = context.socket(zmq.PAIR)
+    def _create_out_control_socket(self, socket_factory: SocketFactory) -> str:
+        self._out_control_socket: Socket = socket_factory.create_socket(SocketType.PAIR)
         out_control_socket_address = f"inproc://BackgroundListener_out_control_socket{id(self)}"
         self._out_control_socket.bind(out_control_socket_address)
         return out_control_socket_address
@@ -63,7 +64,7 @@ class BackgroundListenerInterface:
     def _set_my_connection_info(self):
         message = None
         try:
-            message = self._out_control_socket.recv()
+            message = self._out_control_socket.receive()
             message_obj: Message = deserialize_message(message, Message)
             specific_message_obj = message_obj.__root__
             assert isinstance(specific_message_obj, MyConnectionInfoMessage)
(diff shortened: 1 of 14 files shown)


exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_listener_interface.py
```diff
@@ -48,22 +49,22 @@ def __init__(self,
...
         self._in_control_socket.bind(in_control_socket_address)
         return in_control_socket_address
 
-    def _create_out_control_socket(self, context) -> str:
-        self._out_control_socket: zmq.Socket = context.socket(zmq.PAIR)
+    def _create_out_control_socket(self, socket_factory: SocketFactory) -> str:
+        self._out_control_socket: Socket = socket_factory.create_socket(SocketType.PAIR)
         out_control_socket_address = f"inproc://BackgroundListener_out_control_socket{id(self)}"
         self._out_control_socket.bind(out_control_socket_address)
         return out_control_socket_address
```
Nicoretti, line 62: :spiral_notepad:  Notes:
* Could me merged into a single function, if direction gets parameterized.
* Returning socket and socket address and doing the assignment(s) at the call site, may make it easier to see/spot the side effects

exasol_advanced_analytics_framework/udf_communication/socket_factory/abstract_socket_factory.py
```diff
@@ -0,0 +1,99 @@
+import abc
+from enum import Enum, auto
+from typing import Union, List, Set, Optional
+
+
+class Frame(abc.ABC):
+    @abc.abstractmethod
+    def to_bytes(self) -> bytes:
+        pass
```
Nicoretti, line 9: why not require `__bytes__` to be implemented?

```python
id_frame = IdFrame(id=1, data="some data")

raw_data = bytes(id_frame)
socket.send(raw_data)

```

exasol_advanced_analytics_framework/udf_communication/socket_factory/abstract_socket_factory.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+class PollerFlag(Enum):
+    POLLIN = auto()
+    POLLOUT = auto()
+
+
+class Socket(abc.ABC):
+    @abc.abstractmethod
+    def send(self, message: bytes):
```
Nicoretti, line 19: if you would require anything that's byte convertible (implements `__bytes__`) you could just pass frames or raw data.
e.g.:
```python
class MagicSocket(Socket):

    def send(self, data):
        self.sock.send(bytes(data))
    ...

frame = IdFrame(id=1, data='some text')
sock = MagicSocket()

sock.send(frame)
sock.send(b"binary data")
...

```

exasol_advanced_analytics_framework/udf_communication/socket_factory/zmq_socket_factory.py
```diff
@@ -0,0 +1,145 @@
+from typing import Union, List, Set, Optional, Dict
+
+import zmq
+
+from exasol_advanced_analytics_framework.udf_communication.socket_factory.abstract_socket_factory import Frame, Socket, \
+    PollerFlag, Poller, SocketFactory, SocketType
+
+
+def _flags_to_bitmask(flags: Union[PollerFlag, Set[PollerFlag]]) -> int:
```
Nicoretti, line 9: You may want to use the [IntEnum](https://docs.python.org/3/library/enum.html?highlight=intenum#enum.IntEnum) or [Flag](https://docs.python.org/3/library/enum.html?highlight=intenum#enum.Flag)-Enum for the flag classes.

tests/udf_communication/socket_factory/test_zmq_poller.py
```diff
@@ -0,0 +1,31 @@
...
+            poller = factory.create_poller()
+            poller.register(socket1, flags=PollerFlag.POLLIN)
+            poller.register(socket2, flags=PollerFlag.POLLIN)
+            poller.register(socket3, flags=PollerFlag.POLLIN)
+            poller.register(socket4, flags=PollerFlag.POLLIN)
+            result = poller.poll()
+            assert result == {
+                socket2: {PollerFlag.POLLIN},
+                socket4: {PollerFlag.POLLIN}
+            }
```
Nicoretti, line 31: With an [ExitStack](https://docs.python.org/3/library/contextlib.html?highlight=context%20stack#contextlib.ExitStack) you may be able to simplify or remove some repition.

sudo code:

```python
def test_...(addresses):
    ...
    factory = ZMQSocketFactory(context)
    with ExitStack() as stack:
        sockets = [stack.enter_context(factory.create_socket(SocketType.PAIR)) for _ in addresses]
        ...

```

tests/udf_communication/test_local_discovery.py
```diff
@@ -34,12 +36,15 @@
 
 def run(name: str, group_identifier: str, number_of_instances: int, queue: BidirectionalQueue):
     local_discovery_socket = LocalDiscoverySocket(Port(port=44444))
+    listen_ip = IPAddress(ip_address=f"127.1.0.1")
```
Nicoretti, line 39:
```suggestion
    listen_ip = IPAddress(ip_address="127.1.0.1")
```

exasol_advanced_analytics_framework/udf_communication/socket_factory/abstract_socket_factory.py
tkilias: The ZMQ-Frame is a bit special, it is actually a placeholder which keeps a memory buffer in C alive, and my abstraction follows this.

exasol_advanced_analytics_framework/udf_communication/socket_factory/abstract_socket_factory.py
tkilias: I am using different API for Frame and binary, because the underlying API distinguish them as well.

exasol_advanced_analytics_framework/udf_communication/socket_factory/zmq_socket_factory.py
tkilias: I actually, decided against it, because I didn't want to use bitsets, because in my opinion they are hard to use correctly.

tests/udf_communication/socket_factory/test_zmq_poller.py
tkilias: These sockets are not a file descriptor, or at least not only, you might produce inconsistent object with your example

exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_listener_interface.py
tkilias: This I need to check in depth.

tests/udf_communication/socket_factory/test_zmq_poller.py
tkilias: ah, ok yeah this could work, understood it incorrectly first.

tkilias: But, to be honest, it introduces for 4 sockets a mechanic you need to understand. If we had more sockets, maybe 10, I would agree.

exasol_advanced_analytics_framework/udf_communication/socket_factory/abstract_socket_factory.py
Nicoretti: you still could keep the buffer alive, but anyway was just an idea.

Nicoretti (dismissed):

Nicoretti (dismissed):

exasol_advanced_analytics_framework/udf_communication/socket_factory/zmq_socket_factory.py
```diff
@@ -96,9 +96,10 @@ def __enter__(self):
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        self.close()
+        self.close(linger=0)
 
     def __del__(self):
+        self.close(linger=0)
```
Nicoretti, line 102: As discussed, you may want to add a resource warning here.

Nicoretti (approved):

Nicoretti (approved):

Nicoretti (approved):

PR #123, opened 2023-08-25
#122:  Add new process for stopping a PeerCommunicator 

### LINKED ISSUE #122: Add new process for stopping a PeerCommunicator
# Background

- Currently, we are stopping PeerCommunicator without waiting for the BackgroundThread, directly after the user code finishes successful
- This can lead to issues, because the BackgroundThread might need to send Acks to other Peers
- If it doesn't send these Acks, the other Peer might get stuck, until a timeout kills it
- For that reason, the PeerCommuincator should close in 3 phases
  1. Prepare Stop: In this phase, we announce to the BackgroundThread, we would like to stop, this allows the BackgroundThread to block new incomming tasks and finish the current active ones
  2. Wait: We wait in the PeerCommunicator for the ready to stop signal from the BackgroundThread, which it should send after it has all active tasks finished
  3. Stop: Now we are actually stopping the PeerCommunicator and the BackgroundThread
- If the BackgroundThread doesn't signal the ready to stop after a certain time, we still forcefully stop the PeerCommunicator and the BackgroundThread
```[tasklist]
### Dependencies
- [ ] https://github.com/exasol/advanced-analytics-framework/issues/126
- [ ] https://github.com/exasol/advanced-analytics-framework/issues/124
```

tkilias:
Fixes #122


exasol_advanced_analytics_framework/udf_communication/peer_communicator/acknowledge_register_peer_sender.py
```diff
@@ -56,6 +56,11 @@ def _should_we_send(self):
         result = is_time and not self._finished
         return result
 
+    def is_ready_to_stop(self):
+        result = (self._finished and self._needs_to_send_for_peer) or not self._needs_to_send_for_peer
```
tomuben, line 60: as logically the same as
```suggestion
        result = self._finished or not self._needs_to_send_for_peer
```

exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_listener_thread.py
```diff
@@ -112,16 +116,16 @@ def run(self):
         try:
             self._run_message_loop()
         finally:
-            self._close()
+            self._stop()
 
-    def _close(self):
+    def _stop(self):
         self._logger.info("start")
```
tomuben, line 122: start?

exasol_advanced_analytics_framework/udf_communication/peer_communicator/connection_establisher.py
```diff
@@ -84,3 +84,15 @@ def try_send(self):
         self._abort_timeout_sender.try_send()
         self._peer_is_ready_sender.try_send()
         self._acknowledge_register_peer_sender.try_send()
+
+    def is_ready_to_stop(self):
```
tomuben, line 88: Add type hint for return type

exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_peer_state.py
```diff
@@ -61,5 +61,8 @@ def received_register_peer_complete(self):
     def forward_payload(self, frames: List[Frame]):
         self._receive_socket.send_multipart(frames)
 
-    def close(self):
+    def stop(self):
         self._receive_socket.close(linger=0)
+
+    def is_ready_to_stop(self):
```
tomuben, line 67: add type hint for return type

exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_listener_thread.py
```diff
@@ -146,31 +150,51 @@ def _create_poller(self):
...
+                self._check_is_ready_to_stop()
         except Exception as e:
             self._logger.exception("Exception in message loop")
 
+    def _check_is_ready_to_stop(self):
+        if self._status == BackgroundListenerThread.Status.PREPARE_TO_STOP:
+            if self._is_ready_to_stop():
+                self._out_control_socket.send(serialize_message(IsReadyToStop()))
+
+    def _is_ready_to_stop(self):
```
tomuben, line 165: add type hint for return type

exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_listener_thread.py
```diff
@@ -146,31 +150,51 @@ def _create_poller(self):
...
+        peers_status = [peer_state.is_ready_to_stop()
+                        for peer_state in self._peer_state.values()]
+        is_ready_to_stop = all(peers_status) and len(peers_status) == self._number_of_peers - 1
```
tomuben, line 168: How  could ` len(peers_status) != self._number_of_peers - 1` happen ? Consequences would be that it never reaches ready_to_stop, right?

exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_peer_state.py
```diff
@@ -42,7 +42,7 @@ def _create_receive_socket(self):
         receive_socket_address = get_peer_receive_socket_name(self._peer)
         self._receive_socket.bind(receive_socket_address)
 
-    def resend_if_necessary(self):
+    def try_send(self):
         self._logger.debug("resend_if_necessary")
```
tomuben, line 46:
```suggestion
    def try_send(self):
        self._logger.debug("try_send")
```

exasol_advanced_analytics_framework/udf_communication/peer_communicator/peer_is_ready_sender.py
```diff
@@ -116,6 +116,9 @@ def _send_peer_is_ready_to_frontend(self):
         serialized_message = serialize_message(message)
         self._out_control_socket.send(serialized_message)
 
+    def is_ready_to_stop(self):
```
tomuben, line 119: add type hint for return value

exasol_advanced_analytics_framework/udf_communication/peer_communicator/register_peer_sender.py
```diff
@@ -56,6 +56,9 @@ def _should_we_send(self):
         result = is_time and not self._finished
         return result
 
+    def is_ready_to_stop(self):
```
tomuben, line 59: add type hint for return value

exasol_advanced_analytics_framework/udf_communication/peer_communicator/register_peer_sender.py
```diff
@@ -56,6 +56,9 @@ def _should_we_send(self):
...
+        return (self._finished and self._needs_to_send_for_peer) or not self._needs_to_send_for_peer
```
tomuben, line 60: same as above, can be simplified to:
```suggestion
        return self._finished or not self._needs_to_send_for_peer
```

exasol_advanced_analytics_framework/udf_communication/peer_communicator/peer_communicator_config.py
```diff
@@ -12,3 +12,4 @@ class PeerCommunicatorConfig:
     forward_register_peer_config: ForwardRegisterPeerConfig = ForwardRegisterPeerConfig()
     poll_timeout_in_ms: int = 200
     send_socket_linger_time_in_ms: int = 100
+    close_timeout_in_ms: int = 100000
```
tomuben, line 15: I guess this is an empiric value, and we could decrease later, if to big?

tests/integration_tests/without_db/udf_communication/peer_communication/test_add_peer_forward_and_close.py
```diff
@@ -0,0 +1,158 @@
...
+    logger = LOGGER.bind(group_identifier=parameter.group_identifier, name=parameter.instance_name)
+    try:
+        listen_ip = IPAddress(ip_address=f"127.1.0.1")
+        context = zmq.Context()
+        socket_factory = ZMQSocketFactory(context)
+        socket_factory = FaultInjectionSocketFactory(socket_factory, 0.01, RandomState(parameter.seed))
+        leader = False
+        leader_name = "i0"
+        if parameter.instance_name == leader_name:
+            leader = True
```
tomuben, line 58:
```suggestion
        leader_name = "i0"
        leader = True if parameter.instance_name == leader_name else False
```

tests/integration_tests/without_db/udf_communication/peer_communication/test_add_peer_forward_and_close.py
```diff
@@ -0,0 +1,158 @@
...
+                    is_leader=leader,
+                    is_enabled=True
+                ),
+            ),
+            socket_factory=socket_factory
+        )
+        try:
+            queue.put(com.my_connection_info)
+            peer_connection_infos = queue.get()
+            if parameter.instance_name == leader_name:
```
tomuben, line 75:
```suggestion
            if leader:
```

exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_listener_thread.py
tkilias: the "start" is the event of this logger call, the logging adds module and function name automatically

exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_listener_thread.py
tkilias: This is a common misunderstanding. Each peer stores its state in relation to each other peer. That means we only have a peer_state for the other peers, but not for us self. With that, we only have self._number_of_peers-1 many peers_states.

exasol_advanced_analytics_framework/udf_communication/peer_communicator/peer_communicator_config.py
tkilias: It is more or less arbitrary. It is in the range of the abort timeout, which is also somewhat arbitrary. The idea is, waiting 1–2 minutes is not an issue.

exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_listener_thread.py
```diff
@@ -146,31 +150,51 @@ def _create_poller(self):
...
+    def _try_send(self):
+        if self._status != BackgroundListenerThread.Status.STOPPED:
+            for peer_state in self._peer_state.values():
+                peer_state.try_send()
+
+    def _handle_message(self):
+        poll = self.poller.poll(timeout_in_ms=self._config.poll_timeout_in_ms)
+        if self._in_control_socket in poll and PollerFlag.POLLIN in poll[self._in_control_socket]:
+            message = self._in_control_socket.receive()
+            self._status = self._handle_control_message(message)
```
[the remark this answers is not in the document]
tkilias: A risk, yes, but in the end depends on the protocol which is implemented further down.

exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_listener_thread.py
tomuben: My question is more related to the case when a peer might be down or has nw issues or whatever, then the background_listener_thread never reach ready_to_stop. If this is intended?

tkilias: Ah, now I understand the question. Yes, this is intended. When this happens, the stop timeout sets in and stop everything never the less, but ends with an exception. The reason, for this behavior is, if we can't successful stop everything it is likely, that one of the other peers is not reachable or stuck, in this case we can't successful end, but we don't want to wait forever either.

exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_listener_thread.py
```diff
@@ -146,45 +150,65 @@ def _create_poller(self):
...
                     self._handle_register_peer_message(specific_message_obj)
                 else:
-                    self._logger.error("RegisterPeerMessage message not allowed",
+                    self._logger.error("RegisterPeer message not allowed",
                                        message_obj=specific_message_obj.dict())
             else:
                 self._logger.error("Unknown message type", message_obj=specific_message_obj.dict())
         except Exception as e:
             self._logger.exception("Exception during handling message", message=message)
         return BackgroundListenerThread.Status.RUNNING
```
tkilias, line 203:
```suggestion
        return self._status
```

PR #127, opened 2023-08-25
#126: Split GitHub workflow for tests into separate workflows for each test type

### LINKED ISSUE #126: Split GitHub workflow for tests into separate workflows for each test type
# Background

- We have three test types:
  - unit tests
  - integration tests without DB
  - integration tests with DB
- The different types of tests need different setup
- The setup can be expensive and the integration tests can be expensive, so it might make sense to run the different test types in parallel 
```[tasklist]
### Dependencies
- [ ] https://github.com/exasol/advanced-analytics-framework/issues/124
```

tkilias:
Fixes #126

### DIFF
--- a/.github/actions/prepare_poetry_env/action.yml
+++ b/.github/actions/prepare_poetry_env/action.yml
@@ -11,9 +11,11 @@ runs:
     - uses: actions/setup-python@v2
       with:
         python-version: ${{ inputs.python-version }}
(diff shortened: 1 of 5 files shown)


MarleneKress79789 (approved):
looks good to me

MarleneKress79789 (approved):

PR #156, opened 2023-09-06
#154: Move sending messages to peers to the BackgroundPeerState

### LINKED ISSUE #154: Move sending messages to peers to the BackgroundPeerState
Currently, the FrontendPeerState sends the message to another peer directly via zmq. For the introduction of the PayloadHandler, which handles also repeats and acknowledgements, we need to move this to the BackgroundPeerState , because the PayloadHandler will run in the Background.

tkilias:
Fixes #154

### DIFF
--- a/exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_listener_interface.py
+++ b/exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_listener_interface.py
@@ -94,8 +94,8 @@ class BackgroundListenerInterface:
         self._in_control_socket.send(serialize_message(register_message))
 
     def send_payload(self, message: messages.Payload, payload: List[Frame]):
-        message = serialize_message(message)
-        frame = self._socket_factory.create_frame(message)
+        serialized_message = serialize_message(message)
+        frame = self._socket_factory.create_frame(serialized_message)
         self._in_control_socket.send_multipart([frame] + payload)
 
     def receive_messages(self, timeout_in_milliseconds: Optional[int] = 0) -> Iterator[messages.Message]:
--- a/exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_listener_interface.py
+++ b/exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_listener_interface.py
@@ -1,6 +1,6 @@
 import threading
 from dataclasses import asdict
-from typing import Optional, Iterator
+from typing import Optional, Iterator, List
 
 import structlog
 from structlog.types import FilteringBoundLogger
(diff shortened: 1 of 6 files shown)


exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_listener_interface.py
```diff
@@ -93,6 +93,11 @@ def register_peer(self, peer: Peer):
         register_message = messages.RegisterPeer(peer=peer)
         self._in_control_socket.send(serialize_message(register_message))
 
+    def send_payload(self, message: messages.Payload, payload: List[Frame]):
+        message = serialize_message(message)
```
tomuben, line 97: Do you intentionally overwrite parameter `message` here?

exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_listener_thread.py
```diff
@@ -158,15 +158,15 @@ def _try_send(self):
...
         if self._in_control_socket in poll and PollerFlag.POLLIN in poll[self._in_control_socket]:
-            message = self._in_control_socket.receive()
+            message = self._in_control_socket.receive_multipart()
             self._status = self._handle_control_message(message)
         if self._listener_socket in poll and PollerFlag.POLLIN in poll[self._listener_socket]:
             message = self._listener_socket.receive_multipart()
             self._handle_listener_message(message)
 
-    def _handle_control_message(self, message: bytes) -> Status:
+    def _handle_control_message(self, message: List[Frame]) -> Status:
```
tomuben, line 167: maybe better
```suggestion
    def _handle_control_message(self, frames: List[Frame]) -> Status:
```

exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_listener_interface.py
```diff
@@ -93,6 +93,11 @@ def register_peer(self, peer: Peer):
...
+        frame = self._socket_factory.create_frame(message)
+        self._in_control_socket.send_multipart([frame] + payload)
```
tomuben, line 99: I'am a bit confused: I thought `out_control_socket` was for sending message within host between layers, and `in_control_socket` was for receiving messages within layers. Why do you now send to `in_control_socket`?

tomuben:
Based on the title of the PR I would have expected the send implementation now in `BackgroundPeerState`, but in fact it has been moved to `BackgroundListenerInterface`

exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_listener_interface.py
tkilias: Not really, will use a different variable

exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_listener_interface.py
tkilias: Opposite of the same medal. This class lives on the Frontend and sends to the Backend. We talked until now only about the Backend. However, you are right the names are confusing. I think, we will need to inject also here a BackendSender and a BackendReciever to make this more clear and make this class also unit testable.

tkilias:
> Based on the title of the PR I would have expected the send implementation now in `BackgroundPeerState`, but in fact it has been moved to `BackgroundListenerInterface`

You completely confused me, because I didn't move the receive in #157 completely to the BackgroundPeerState, but in fact the send went to the BackgroundPeerState.

tomuben (approved):

PR #159, opened 2023-09-08
#152: Use payload handler for sending and receiving

### LINKED ISSUE #152: Use PayloadHandler for sending and receiving messages
- PeerCommunicator _handle_messages needs to handle AcknowledgePayload Messages
  - AcknowledgePayload get for the moment ignored, in a later PR, we can use this for MessageTracking
-  BackgroundPeerState has to use the PayloadHandler for handling Payload related messages
```[tasklist]
### Dependencies
- [ ] https://github.com/exasol/advanced-analytics-framework/issues/151
- [ ] https://github.com/exasol/advanced-analytics-framework/issues/155
- [ ] https://github.com/exasol/advanced-analytics-framework/issues/154
- [ ] https://github.com/exasol/advanced-analytics-framework/issues/139
```

tkilias:
Fixes #152

### DIFF
--- a/exasol_advanced_analytics_framework/udf_communication/messages.py
+++ b/exasol_advanced_analytics_framework/udf_communication/messages.py
@@ -61,6 +61,7 @@ class Payload(BaseMessage, frozen=True):
 class AcknowledgePayload(BaseMessage, frozen=True):
     message_type: Literal["AcknowledgePayload"] = "AcknowledgePayload"
     source: Peer
+    destination: Peer
     sequence_number: int
 
 
@@ -79,6 +80,7 @@ class SynchronizeConnection(BaseMessage, frozen=True):
     message_type: Literal["SynchronizeConnection"] = "SynchronizeConnection"
     source: ConnectionInfo
     destination: Peer
+    attempt: int
 
 
 class AcknowledgeConnection(BaseMessage, frozen=True):
@@ -96,6 +98,7 @@ class CloseConnection(BaseMessage, frozen=True):
     message_type: Literal["CloseConnection"] = "CloseConnection"
     source: ConnectionInfo
     destination: Peer
+    attempt: int
 
 
 class AcknowledgeCloseConnection(BaseMessage, frozen=True):
--- a/exasol_advanced_analytics_framework/udf_communication/peer_communicator/frontend_peer_state.py
+++ b/exasol_advanced_analytics_framework/udf_communication/peer_communicator/frontend_peer_state.py
@@ -81,4 +81,3 @@ class FrontendPeerState:
 
     def received_acknowledge_payload_message(self, acknowledge_payload: messages.AcknowledgePayload):
         """ Not yet implemented and for that reason we ignore the input"""
-        pass
--- a/exasol_advanced_analytics_framework/udf_communication/peer_communicator/frontend_peer_state.py
+++ b/exasol_advanced_analytics_framework/udf_communication/peer_communicator/frontend_peer_state.py
@@ -78,3 +78,7 @@ class FrontendPeerState:
     @property
     def connection_is_closed(self) -> bool:
         return self._connection_is_closed
+
+    def received_acknowledge_payload_message(self, acknowledge_payload: messages.AcknowledgePayload):
+        """ Not yet implemented and for that reason we ignore the input"""
+        pass
--- a/exasol_advanced_analytics_framework/udf_communication/peer_communicator/peer_communicator.py
+++ b/exasol_advanced_analytics_framework/udf_communication/peer_communicator/peer_communicator.py
@@ -92,6 +92,11 @@ class PeerCommunicator:
                 raise TimeoutError(specific_message_obj.reason)
             elif isinstance(specific_message_obj, messages.Payload):
                 self._peer_states[specific_message_obj.source].received_payload_message(specific_message_obj, frames)
+            elif isinstance(specific_message_obj, messages.AcknowledgePayload):
+                self._peer_states[specific_message_obj.source].received_acknowledge_payload_message(
+                    specific_message_obj)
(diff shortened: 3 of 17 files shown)


exasol_advanced_analytics_framework/udf_communication/messages.py
```diff
@@ -79,6 +80,7 @@ class SynchronizeConnection(BaseMessage, frozen=True):
     message_type: Literal["SynchronizeConnection"] = "SynchronizeConnection"
     source: ConnectionInfo
     destination: Peer
+    attempt: int
```
tomuben, line 83: Maybe also rename it like in exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_thread/connection_closer/close_connection_sender.py
`attempt_count` or `attempt_seq_number` or `attempt_idx`

exasol_advanced_analytics_framework/udf_communication/peer_communicator/frontend_peer_state.py
```diff
@@ -78,3 +78,7 @@ def received_connection_is_closed(self):
     @property
     def connection_is_closed(self) -> bool:
         return self._connection_is_closed
+
+    def received_acknowledge_payload_message(self, acknowledge_payload: messages.AcknowledgePayload):
+        """ Not yet implemented and for that reason we ignore the input"""
+        pass
```
tomuben, line 84: I learned from Nico that if you have a docstring you don't even need to declare "pass" here :)

exasol_advanced_analytics_framework/udf_communication/peer_communicator/peer_communicator.py
```diff
@@ -92,6 +92,11 @@ def _handle_messages(self, timeout_in_milliseconds: Optional[int] = 0):
                 raise TimeoutError(specific_message_obj.reason)
             elif isinstance(specific_message_obj, messages.Payload):
                 self._peer_states[specific_message_obj.source].received_payload_message(specific_message_obj, frames)
+            elif isinstance(specific_message_obj, messages.AcknowledgePayload):
+                self._peer_states[specific_message_obj.source].received_acknowledge_payload_message(
+                    specific_message_obj)
+            elif isinstance(specific_message_obj, messages.AbortPayload):
+                raise TimeoutError(specific_message_obj.reason)
```
tomuben, line 99: Is it enough to only raise the exception in this case? No buffer cleanup needed?

exasol_advanced_analytics_framework/udf_communication/socket_factory/fault_injection.py
```diff
@@ -47,7 +50,7 @@ def _is_fault(self):
 
     def send(self, message: bytes):
         if self._is_fault():
-            self._logger.info("Fault injected", message=message)
+            self._logger.warning("Fault injected", message=message)
```
tomuben, line 53: Is this only for debugging or a permanent change?

tests/unit_tests/udf_communication/peer_communication/test_payload_receiver.py
```diff
@@ -168,6 +170,21 @@ def test_is_ready_to_stop_after_received_payload_in_reverse_sequence(number_of_m
...
+@pytest.mark.parametrize("number_of_messages, duplicated_message",
+                         [(i, j) for i in range(1, 10) for j in range(0, i)])
+def test_is_ready_to_stop_payload_in_sequence_multiple_times(number_of_messages: int, duplicated_message: int):
+    test_setup = create_test_setup()
+    for sequence_number in range(number_of_messages):
+        message, frames = create_payload_message(test_setup, sequence_number)
+        test_setup.payload_receiver.received_payload(message, frames)
+    test_setup.reset_mock()
+    sequence_number = duplicated_message
+    message, frames = create_payload_message(test_setup, sequence_number)
```
tomuben, line 182: You only create a single "duplicated" message per test. 
Wouldn't it be better to create all duplicated messages here?

exasol_advanced_analytics_framework/udf_communication/socket_factory/fault_injection.py
tkilias: I want to make it a permanent change, because this is anyway for testing only and you should at least know which message were lost by the fault injector

tests/unit_tests/udf_communication/peer_communication/test_payload_receiver.py
tkilias: Maybe, for the case I wanted to test, a single duplicated is enough.

exasol_advanced_analytics_framework/udf_communication/peer_communicator/peer_communicator.py
tkilias: As a user you should not catch the exceptions of the peer_communicator. I thought about to do a hard cleanup. Maybe as an additional PR.

exasol_advanced_analytics_framework/udf_communication/peer_communicator/frontend_peer_state.py
tkilias: The pass first, yeah otherwise you are right, I forgot it.

exasol_advanced_analytics_framework/udf_communication/messages.py
tkilias: Hm, attempt seems perfect for me. Count is something absolute, so this doesn't fit. It is also not really an idx number. Sequence number would most, but has usually also different meaning.

tests/unit_tests/udf_communication/peer_communication/test_payload_receiver.py
tkilias: Changed it to duplicate all until duplicate idx

tomuben (approved):

PR #161, opened 2023-09-11
#160: Implement ConnectionCloser

### LINKED ISSUE #160: Implement ConnectionCloser
# Background

- When implementing #152, we recognized that peers can get stuck when the last Payload or PayloadAck gets lost.
- The reason is, that one side assumes too early it is ready and stops
- This causes the PayloadSender never to receive a PayloadAck
- With that the PayloadSender is never ready to stop. Thus, the Peer can't stop gracefully
- Another, scenario that can happen is that on loaded systems the Peers make progress at different speeds, which can lead to that one Peer is much earlier ready than another one.
- This can lead basically to a similar scenario where one of the Peers can't gracefully stop
- We can fix this with a 4 way handshake, where the last ACK can get lost. However, the Peer which is waiting for the ACK than needs to wait a certain time
  - This is similar to the tcp close except that in our case both Peers send the FIN message independent of each other and the peer which did yet receive a FIN from the other node needs to stay alive and potentially receive messages and react on them
    - https://en.wikipedia.org/wiki/Transmission_Control_Protocol#Connection_termination 
- The following diagram shows the event sequence 

connection_finish_sequence drawio

# Acceptance Criteria

- ConnectionCloser implementation
- Unit tests for it

# Out of scope

- Integration into BackgroundPeerState

tkilias:
Fixes #160

### DIFF
--- a/exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_thread/connection_closer/connection_is_closed_sender.py
+++ b/exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_thread/connection_closer/connection_is_closed_sender.py
@@ -1,0 +1,97 @@
+from enum import IntFlag, auto
+
+import structlog
+from structlog.typing import FilteringBoundLogger
+
+from exasol_advanced_analytics_framework.udf_communication import messages
+from exasol_advanced_analytics_framework.udf_communication.connection_info import ConnectionInfo
+from exasol_advanced_analytics_framework.udf_communication.peer import Peer
+from exasol_advanced_analytics_framework.udf_communication.peer_communicator.timer import Timer
+from exasol_advanced_analytics_framework.udf_communication.serialization import serialize_message
+from exasol_advanced_analytics_framework.udf_communication.socket_factory.abstract import Socket
+
+LOGGER: FilteringBoundLogger = structlog.get_logger()
+
+
+class _States(IntFlag):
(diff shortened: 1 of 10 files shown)


exasol_advanced_analytics_framework/udf_communication/peer_communicator/background_thread/connection_closer/connection_is_closed_sender.py
```diff
@@ -0,0 +1,97 @@
...
+        should_we_send = self._should_we_send()
+        if should_we_send:
+            self._states |= _States.FINISHED
+            self._send_connection_is_closed_to_frontend()
+
+    def _should_we_send(self):
+        is_time = self._timer.is_time()
+        send_time_dependent = _States.RECEIVED_CLOSE_CONNECTION in self._states
+        send_time_independent = (
+                _States.RECEIVED_CLOSE_CONNECTION in self._states
```
tkilias, line 59: This is the only line that really changed compared to the ConnectionEstablisher, because for the ConnectionEstablisher it only counts if the other Peers does know us. If the other peer receives a Synchronize message, it answers with a Synchronize and a Ack message, so we only need on of it. However, for the ConnectionCloser, we need that both sides send CloseConnection, because otherwise we don't know it the other one wants to still send something to us.

tomuben (approved):

PR #192, opened 2024-10-09
#191: Renamed `parameter` to `parameters` to match JSON input in usrr guide

### LINKED ISSUE #191: Rename UDF json element "parameters" to "parameter"
Currently, the example in the user guide specifies plural `"parameters": "bla-bla"` while the constructor of the example UDF class contains an argument `parameter` (singular). 

In json you probably can specify other datatypes such as int, float or boolean as well as more complex data, such as a json object or an array while the argument of the constructor of the UDF class still remains a string.

Probably, the json value is converted to a string and the UDF constructor can optionally use json library to get back other datatypes. 

However, the argument's numerus should be identical.

The current ticket requests to name it to singular. This requires updating 
* the user guide
* the lua script [query_loop.lua](https://github.com/exasol/advanced-analytics-framework/blob/main/exasol_advanced_analytics_framework/lua/src/query_loop.lua).

ckunki:
Closes  #191

### DIFF
--- a/tests/test_package/test_query_handlers/query_handler_test.py
+++ b/tests/test_package/test_query_handlers/query_handler_test.py
@@ -27,12 +27,12 @@ QUERY_LIST = [SelectQuery("SELECT 1 FROM DUAL"), SelectQuery("SELECT 2 FROM DUAL
 
 class QueryHandlerTestWithOneIteration(UDFQueryHandler):
 
-    def __init__(self, parameter: str, query_handler_context: ScopeQueryHandlerContext):
-        super().__init__(parameter, query_handler_context)
-        if not isinstance(parameter, str):
-            raise AssertionError(f"Expected parameter={parameter} to be a string.")
-        if parameter != TEST_INPUT:
-            raise AssertionError(f"Expected parameter={parameter} to be '{TEST_INPUT}'.")
+    def __init__(self, parameters: str, query_handler_context: ScopeQueryHandlerContext):
+        super().__init__(parameters, query_handler_context)
+        if not isinstance(parameters, str):
+            raise AssertionError(f"Expected parameters={parameters} to be a string.")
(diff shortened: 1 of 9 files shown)


tests/test_package/test_query_handlers/query_handler_test.py
```diff
@@ -99,14 +99,14 @@ def handle_query_result(self, query_result: QueryResult) -> Union[Continue, Fini
 
 class QueryHandlerWithOneIterationWithNotReleasedChildQueryHandlerContextFactory(UDFQueryHandlerFactory):
 
-    def create(self, parameter: str, query_handler_context: ScopeQueryHandlerContext) -> UDFQueryHandler:
-        return QueryHandlerWithOneIterationWithNotReleasedChildQueryHandlerContext(parameter, query_handler_context)
+    def create(self, parameters: str, query_handler_context: ScopeQueryHandlerContext) -> UDFQueryHandler:
+        return QueryHandlerWithOneIterationWithNotReleasedChildQueryHandlerContext(parameters, query_handler_context)
```
ahsimb, line 103: reminds me a Welsh train station name:
Llanfairpwllgwyngyllgogerychwyrndrobwllllantysiliogogogoch

ahsimb (approved):

tests/test_package/test_query_handlers/query_handler_test.py
tkilias: This is actually a test case, thats why we need the lengthy name to be precise

tkilias (changes requested):
We wanted to go with parameter and not parameters.

tests/test_package/test_query_handlers/query_handler_test.py
ckunki: :rofl:

PR #202, opened 2024-10-21
#190 enabled to generate a dynamic module for custom udf

### LINKED ISSUE #190: Generate dynamic module for custom UDF
Factory on the fly 

See
* https://joeletts.com/2020/creating-dynamic-modules-python-3/
* https://www.b-list.org/weblog/2007/dec/03/making-magic/

Tasks
* [x] Investigate creating dynamic modules
* [x] Use dynamic modules for adhoc implementation of an example Query Handler
* [x] Integration test for example based on dynamic modules

ckunki:
Closes #190


doc/changes/changes_0.1.0.md
```diff
@@ -52,6 +52,7 @@ Code name:
 * #176: Updated usage of `exasol-bucketfs` to new API
 * #185: Removed directory and script for building SLC AAF
 * #191: Renamed UDF json element "parameters" to "parameter"
+* #190: Enabled to generate a dynamic module for custom UDF
```
ahsimb, line 55: Enabled what?

doc/developer_guide/developer_guide.md
```diff
@@ -14,6 +14,28 @@ poetry run nox -s build_language_container
 
 Installing the SLC ins described in the [AAF User Guide](../user_guide/user_guide.md#script-language-container-slc).
 
+## Updated Generated Files
+
+AAF contains some generated files that are committed to git, though:
+* The amalgated Lua script [create_query_loop.sql](https://github.com/exasol/advanced-analytics-framework/blob/main/exasol_advanced_analytics_framework/resources/outputs/create_query_loop.sql)
+* The examples in the user guide
+
+The amalgated Lua script originates from the files in directory [exasol_advanced_analytics_framework/lua/src](https://github.com/exasol/advanced-analytics-framework/blob/main/exasol_advanced_analytics_framework/lua/src/).
```
ahsimb, line 23:
```suggestion
The amalgated Lua script originates from the files in the directory [exasol_advanced_analytics_framework/lua/src](https://github.com/exasol/advanced-analytics-framework/blob/main/exasol_advanced_analytics_framework/lua/src/).
```

doc/developer_guide/developer_guide.md
```diff
@@ -14,6 +14,28 @@ poetry run nox -s build_language_container
...
+
+The following command updates the amalgated script:
+
+```shell
+poetry run nox -s amalgate_lua_scripts
+```
+
+AAF's user guide contains an example for an adhoc implementation of a Query Handler originating from the files in directory [exasol_advanced_analytics_framework/example](https://github.com/exasol/advanced-analytics-framework/blob/main/exasol_advanced_analytics_framework/example/).
```
ahsimb, line 31:
```suggestion
AAF's user guide contains an example of an ad hoc implementation of a Query Handler originating from the files in the directory [exasol_advanced_analytics_framework/example](https://github.com/exasol/advanced-analytics-framework/blob/main/exasol_advanced_analytics_framework/example/).
```

doc/user_guide/user_guide.md
```diff
@@ -188,56 +188,109 @@ Each algorithm should extend the `UDFQueryHandler` abstract class and then imple
 
 ### Concrete Example Using an Adhoc Implementation Within the UDF
 
-The example uses the module `builtins` and dynamically adds `ExampleQueryHandler` and `ExampleQueryHandlerFactory` to it.
+The example dynamically creates a python module `xyz` and adds classes `ExampleQueryHandler` and `ExampleQueryHandlerFactory` to it.
```
ahsimb, line 191: I would give the module a more meaningful name than "xyz". If you were doing this manually you would probably call the module "example_query_handler.py" or "example_qh.py", or something like that.

doc/user_guide/user_guide.md
```diff
@@ -188,56 +188,109 @@ Each algorithm should extend the `UDFQueryHandler` abstract class and then imple
...
 
+In order to execute the example successfully you need to
+1. [Create a BucketFS connection](#bucketfs-connection)
+2. Activate the AAF's SLC
+2. Create the involved database schemas
```
ahsimb, line 196:
```suggestion
3. Make sure the database schemas used in the example exist.
```

doc/user_guide/user_guide.md
```diff
@@ -188,56 +188,109 @@ Each algorithm should extend the `UDFQueryHandler` abstract class and then imple
...
+
+The example assumes the name for the BucketFS Connection `<CONNECTION_NAME>` to be `BFS_CON`.
+
+The following SQL statements activate the AAF's SLC and create the involved database schemas:
```
ahsimb, line 200:
```suggestion
The following SQL statements activate the AAF's SLC and create the required database schemas unless they already exist:
```

noxfile.py
```diff
@@ -52,6 +53,12 @@ def install_dev_env(session: Session):
     session.run(str(install_script))
 
 
+@nox.session(python=False)
+def amalgate_lua_scripts(session: Session):
+    script = ROOT_DIR / "exasol_advanced_analytics_framework" / "deployment" / "regenerate_scripts.py"
+    _run_in_dev_env_poetry_call(session, "python", str(script))
```
Nicoretti, line 59: what is this for, could a nox based venv be a solution here?

noxfile.py
```diff
@@ -124,3 +131,11 @@ def run_python_integration_tests_with_db(session: Session):
...
         *session.posargs,
     )
+
+
+@nox.session(python=False)
+def update_user_guide(session: Session):
+    """
+    This task updates the examples in the user guide.
+    """
+    update_examples(Path("doc/user_guide/user_guide.md"))
```
Nicoretti, line 141: should this be plugged in to the prepare-release hooks to?

exasol_advanced_analytics_framework/example/query_handler.py
```diff
@@ -0,0 +1,82 @@
...
+        c2 = query_result.c2
+        bfs_content = as_string(self._bfs_file(self.bfs_proxy).read())
+        return Finish(result=f"Final result: from query '{c1}', {c2} and bucketfs: '{bfs_content}'")
+
+
+xyz.add_to_module(ExampleQueryHandler)
+
+class ExampleQueryHandlerFactory:
+    def create(self, parameter: str, query_handler_context: QueryHandlerContext):
+        return xyz.ExampleQueryHandler(parameter, query_handler_context)
```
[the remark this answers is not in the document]
ckunki: This question is maybe more related to the AAF API and its intended usage in general.
Maybe you could ask @tkilias ?

exasol_advanced_analytics_framework/example/query_handler.py
```diff
@@ -0,0 +1,82 @@
...
+
+xyz.add_to_module(ExampleQueryHandlerFactory)
+
+from exasol_advanced_analytics_framework.udf_framework.query_handler_runner_udf \
+    import QueryHandlerRunnerUDF
```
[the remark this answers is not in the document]
ckunki: I agree very much. 
This pattern is present in many other repos, too. 
As renaming the modules is currently out of scope or would make the PR too huge: Would you be OK creating a separate ticekt for that?

noxfile.py
ckunki: We need to consider the Lua installation and conda being involved. 
Potentially, this might be possible. 
Not sure. 

However, these parts are currently unchanged and only wrapped into nox tasks in some crucial places. 
In order to not merge too many changes into a single PR I propose to create a separate ticket. 
Could you agree to that?

noxfile.py
ckunki: Currently, it's separate.
There is a pre-commit hook and a github workflow verifying the user guide to be up-to-date. 
So moving it into the prepare-release is not required in my eyes. 
But in general, we could evaluate this idea - it sounds neither bad nor hard to accomplish to me.

scripts/document_updater.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+
+
+@dataclass
+class Template:
+    path: str
+
+    def render(self):
+        if self.path != "example/generator.py":
+            raise ParseException("document_updater.Template currently only"
```
ahsimb, line 24: According to its docstring, the ParseException is supposed to be raised if the file is not well-formed. Maybe ValueError will be a better choice here.

scripts/document_updater.py
```diff
@@ -0,0 +1,97 @@
...
+            "",
+            example_generator.execute_script() + ";",
+            "```",
+            ""])
+
+
+class ChunkReader:
+    """
+    Enables to replace chunks of a string by generated text, e.g. from
+    jinja templates.
```
ahsimb, line 41: The class name and the description do not provide a good explanation of what this class is doing. Maybe a usage example can help.

scripts/document_updater.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def _end_generated(self, line: str):
+        if not self._generated:
+            raise ParseException(
+                f"Found {line} before any <!-- generated from/by ... -->."
+            )
+        self._generated = None
+        self._plain.append(line)
+
+    def split(self, content: str) -> List[str|Template]:
```
ahsimb, line 74: A docstring here would really help explain what this class is doing.

scripts/document_updater.py
```diff
@@ -0,0 +1,97 @@
...
+                self._start_generated(line, match)
+            elif end.match(line):
+                self._end_generated(line)
+            elif not self._generated:
+                self._plain.append(line)
+        self._process_plain()
+        return self._chunks
+
+    @classmethod
+    def chunks(cls, content: str):
```
ahsimb, line 89: Question:
Can I use this class like this
```
my_chunk_reader = ChunkReader()
my_split_content1 = my_chunk_reader.split(my_content1)
my_split_content2 = my_chunk_reader.split(my_content2)
...
```
Probably not, right? But you need to look at the code to realize this is not a good idea.

doc/user_guide/user_guide.md
```diff
@@ -188,56 +188,109 @@ Each algorithm should extend the `UDFQueryHandler` abstract class and then imple
...
+        def table_query(statement: str, **kwargs):
+            return SelectQuery(table_query_string(statement, **kwargs))
+
+        self.bfs_proxy = self.query_handler_context.get_temporary_bucketfs_location()
+        self._bfs_file(self.bfs_proxy).write(sample_content("bucketfs"))
+        self.db_table_proxy = self.query_handler_context.get_temporary_table_name()
         query_list = [
-          SelectQuery("SELECT 1 FROM DUAL"),
-          SelectQuery("SELECT 2 FROM DUAL")]
+            table_query('CREATE TABLE {table_name} ("c1" VARCHAR(100), "c2" INTEGER)'),
```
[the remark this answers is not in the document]
ckunki: The placeholder `{table_name}` is not expanded here, but in function `table_query()`, more precisely in `table_query_string()`: 
```
        def table_query_string(statement: str, **kwargs):
            table_name = self.db_table_proxy._db_object_name.fully_qualified
            return statement.format(table_name=table_name, **kwargs)
```

The name is obtained from the so-called _proxy_.
AAF uses different proxies to enable reference counting and auto-removal when terminating a query handler. 
As the concept seems important too me and your question itches the same point, I will add a short section to the user guide in the next push.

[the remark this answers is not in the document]
ckunki: Additionally: Currently we are abusing the `SelectQuery` also for `CREATE TABLE` and `INSERT` statements as used in the example, as there is no `InsertQuery` or `CreateQuery`, yet.

However, this requires accessing the private name of the database object encapsulated by the proxy as seen in function `table_query_string()` above. Maybe we decide to change this in the current PR already?

scripts/document_updater.py
ckunki: Fine for me, thanks - see next push.

scripts/document_updater.py
ckunki: See next push

scripts/document_updater.py
ckunki: The chunk-reader is only used for updating the examples in the user guide. 
I think you are right, that the sample usage sketched by you would create unexpected results. 
I see 2 options: 
* O1) change the implementation to make such a use impossible, e.g. by using static methods, splitting in the constructor, etc. 
* O2) Enable the reuse of a `ChunkReader` instance as indicated by your example. 

O2 seemed easier to me, so please see next push for an improved version.

Nicoretti (approved):
Further details and discussion can be handled with @ahsimb

scripts/document_updater.py
ckunki: I must also admit that there currently not tests for `ChunkReader`, yet.

I ommitted the tests for the following reasons:
* `ChunkReader` is only used for updating the examples in the user guide
* `ChunkReader` is not mission-critical in my eyes, as you could also update the examples manually as a fall-back
* `ChunkReader` in my eyes could be rated as a kind of build-script similar to the GitHub workflows

For the GitHub workflows there are not tests, either.  On the other hand GitHub workflows as well as the document updater are used in each CI build and hence we have evidence that they are working and potential problems will be detected early.

But if all this doesn't convince you, then maybe there is an even better way to achieve the goals listed in the ticket #190? In this case, I am happy to discuss.

scripts/document_updater.py
ckunki: See next push.

exasol_advanced_analytics_framework/example/query_handler.py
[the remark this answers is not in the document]
ckunki: See #203

scripts/document_updater.py
ahsimb: My solutions to the problem with the design of this class would be one of those:
1. Make the `split` function clean up after itself, so it would be safe to call it multiple times.
2. Make this class private, i.e. `_ChunkReader` and make a function that uses this class like the `update_example`, or maybe the `update_example` itself is all you need.
3. Put it in the class docstring that this class should not be used like in my comment and should be used like in the `update_example`.

ahsimb (dismissed):

exasol_advanced_analytics_framework/udf_framework/dynamic_modules.py
```diff
@@ -0,0 +1,24 @@
...
+def create_module(name: str) -> ModuleType:
+    """
+    Dynamically create a python module using the specified name and
+    register the module in sys.modules[].
+
+    Additionally add a function add_to_module() to the module enabling other
+    code to add classes and functions to the module.
+    """
+    spec = importlib.machinery.ModuleSpec(name, None)
+    mod = importlib.util.module_from_spec(spec)
```
tkilias, line 16: extract to _create_module

exasol_advanced_analytics_framework/udf_framework/dynamic_modules.py
```diff
@@ -0,0 +1,24 @@
...
+    sys.modules[name] = mod
```
tkilias, line 17: extract to _register_module_for_import

tkilias: how do we handle the case if the module already exists?

doc/user_guide/proxies.md
```diff
@@ -0,0 +1,13 @@
+## AAF Proxies
+
+The Advanced Analytics Framework (AAF) uses _Object Proxies_ to manage temporary objects.
+
+An _Object Proxy_
+* Encapsulates a temporary object
+* Provides a reference enabling using the object, i.e. its name incl. the database schema or the path in the BucketFS
+* Ensures the object is removed when leaving the current scope, e.g. the Query Handler. For this the proxy maintains a _reference counter_.
```
tkilias, line 8:
```suggestion
* Ensures the object is removed when leaving the current scope, e.g. the Query Handler..
```

ckunki: Fixed.

scripts/document_updater.py
ckunki: Thanks to @tkilias for inspiring discussion yesterday.

I changed the implementation to 
* go without `ChunkReader`
* without updating the user guide
* replacing the scripts in the user guide by simple references to separate files 
* using these files in the integration test
* only replacing the names of the DB schema and BucketFS connection with `string.replace()`

I think this approach is much simpler, requires less code, no check if the examples in the user guide are out-dated.

Please see latest push.

exasol_advanced_analytics_framework/udf_framework/dynamic_modules.py
ckunki: See next push 
* moving operations to separate functions `_create_module()` and `_register_module_for_import()`
* Covering the case if the module already exists
* Enhancing unit tests to verify the case if the module already exists

ahsimb (approved):

.github/workflows/check-code-generation.yml
tkilias, line 1: Oh, did something like this not already exist, if not this was really an oversight of me.

doc/developer_guide/developer_guide.md
```diff
@@ -14,6 +14,20 @@ poetry run nox -s build_language_container
 
 Installing the SLC ins described in the [AAF User Guide](../user_guide/user_guide.md#script-language-container-slc).
 
+## Update Generated Files
+
+AAF contains some generated files that are committed to git, including:
+* The amalgated Lua script [create_query_loop.sql](https://github.com/exasol/advanced-analytics-framework/blob/main/exasol_advanced_analytics_framework/resources/outputs/create_query_loop.sql)
+* The examples in the user guide
```
tkilias, line 21: is this still true

doc/user_guide/example-udf-script/execute.sql
```diff
@@ -0,0 +1,20 @@
+EXECUTE SCRIPT "AAF_DB_SCHEMA"."AAF_RUN_QUERY_HANDLER"('{
+    "query_handler": {
+        "factory_class": {
+            "module": "example_module",
+            "name": "ExampleQueryHandlerFactory"
+        },
+        "parameter": "bla-bla",
+        "udf": {
+            "schema": "EXAMPLE_SCHEMA",
+            "name": "MY_QUERY_HANDLER_UDF"
```
tkilias, line 10: if we already run with the word example, should we use it everywhere (half a joke)

doc/user_guide/user_guide.md
```diff
@@ -108,10 +108,12 @@ python -m exasol_advanced_analytics_framework.deploy scripts \
     --dsn "$DB_HOST:DB_PORT" \
     --db-user "$DB_USER" \
     --db-pass "$DB_PASSWORD" \
-    --schema "$DB_SCHEMA" \
+    --schema "$AAF_DB_SCHEMA" \
     --language-alias "$LANGUAGE_ALIAS"
 ```
 
+The name of the database schema must match the schema when executing the script.
```
tkilias, line 115: this is unclear

doc/user_guide/user_guide.md
```diff
@@ -188,88 +191,31 @@ Each algorithm should extend the `UDFQueryHandler` abstract class and then imple
...
-      def create(self, parameter: str, query_handler_context: QueryHandlerContext):
-          return builtins.ExampleQueryHandler(parameter, query_handler_context)
+The following SQL statements activate the AAF's SLC and create the required database schemas unless they already exist:
 
-builtins.ExampleQueryHandlerFactory=ExampleQueryHandlerFactory
-
-from exasol_advanced_analytics_framework.udf_framework.query_handler_runner_udf \
-    import QueryHandlerRunnerUDF
+```shell
+ALTER SYSTEM SET SCRIPT_LANGUAGES='R=builtin_r JAVA=builtin_java PYTHON3=builtin_python3 PYTHON3_AAF=localzmq+protobuf:///bfsdefault/default/temp/exasol_advanced_analytics_framework_container_release?lang=python#/buckets/bfsdefault/default/temp/exasol_advanced_analytics_framework_container_release/exaudf/exaudfclient_py3';
```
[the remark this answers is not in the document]
tkilias: and the path actually has many assumptions.

exasol_advanced_analytics_framework/udf_framework/dynamic_modules.py
```diff
@@ -0,0 +1,34 @@
...
+def create_module(name: str) -> ModuleType:
+    """
+    Dynamically create a python module using the specified name and
+    register the module in sys.modules[].
+
+    Additionally add a function add_to_module() to the module enabling other
+    code to add classes and functions to the module.
+    """
+    mod = sys.modules.get(name)
+    if mod is None:
```
tkilias, line 25: I think, we actually should fail, when the module exists

noxfile.py
tkilias: @Nicoretti this is unfortunately a nasty combination out of a conda env, with luarocks env and a poetry env, if you have an idea how we get this done with nox I am happy to hear

exasol_advanced_analytics_framework/udf_framework/dynamic_modules.py
ckunki: See last push

doc/developer_guide/developer_guide.md
ckunki: Thanks - I removed the main instructions but forgot to update the bullet list. See next push.

doc/user_guide/example-udf-script/execute.sql
ckunki: Actually, during my experiments, I switched from MY_SCHEMA to MY_EXAMPLE to avoid false positive test results due to my test database being already partially initialized.

Should I replace the name `MY_QUERY_HANDLER_UDF` by `EXAMPLE_QUERY_HANDLER_UDF`?

doc/user_guide/user_guide.md
ckunki: See next push.

exasol_advanced_analytics_framework/udf_framework/dynamic_modules.py
ckunki: See next push.

doc/user_guide/user_guide.md
```diff
@@ -188,88 +191,31 @@ Each algorithm should extend the `UDFQueryHandler` abstract class and then imple
...
-      def create(self, parameter: str, query_handler_context: QueryHandlerContext):
-          return builtins.ExampleQueryHandler(parameter, query_handler_context)
+The following SQL statements activate the AAF's SLC and create the required database schemas unless they already exist:
 
-builtins.ExampleQueryHandlerFactory=ExampleQueryHandlerFactory
-
-from exasol_advanced_analytics_framework.udf_framework.query_handler_runner_udf \
-    import QueryHandlerRunnerUDF
+```shell
+ALTER SESSION SET SCRIPT_LANGUAGES='R=builtin_r JAVA=builtin_java PYTHON3=builtin_python3 PYTHON3_AAF=localzmq+protobuf:///bfsdefault/default/temp/exasol_advanced_analytics_framework_container_release?lang=python#/buckets/bfsdefault/default/temp/exasol_advanced_analytics_framework_container_release/exaudf/exaudfclient_py3';
```
ckunki, line 208: I applied your proposal, @tkilias.
How about the assumptions? 
Maybe we first try to identify all assumptions?

Thinking about it: maybe we can omit the ALTER SESSION statement, as all this is already done by the AAF deploy command?

tkilias: lets remove it

ckunki: :+1:

tkilias (approved):

PR #205, opened 2024-11-04
Bug/fix download url for slc file

### LINKED ISSUE #206: Fix download URL for SLC file

ckunki:
Closes #206

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -25,12 +25,6 @@ jobs:
       - name: Build Script Language Container (SLC)
         run: poetry run nox -s build_language_container
 
-      - name: PyPi Release
-        env:
-          POETRY_HTTP_BASIC_PYPI_USERNAME: "__token__"
-          POETRY_HTTP_BASIC_PYPI_PASSWORD: "${{ secrets.PYPI_TOKEN }}"
-        run: poetry publish
-
       - name: GitHub Release
         env:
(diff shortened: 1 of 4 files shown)


.github/workflows/build-and-publish.yml
```diff
@@ -25,6 +25,12 @@ jobs:
       - name: Build Script Language Container (SLC)
         run: poetry run nox -s build_language_container
 
+      - name: PyPi Release
+        env:
+          POETRY_HTTP_BASIC_PYPI_USERNAME: "__token__"
+          POETRY_HTTP_BASIC_PYPI_PASSWORD: "${{ secrets.PYPI_TOKEN }}"
+        run: poetry publish
```
Nicoretti, line 32: Is this part of the bugfix?

Nicoretti (dismissed):

.github/workflows/build-and-publish.yml
ckunki: You are right, additionally (as part of the clean-up after the release) I moved the pypi release back to the old position in the workflow, which is before the GitHub release. 
I hope, this is OK for you?

tkilias: But I am not ok with it, because a PyPi release doesn't make sense, if we couldn't upload the SLC

tkilias (changes requested):
Revert the movement of the PyPi release

ckunki:
> Revert the movement of the PyPi release

Reverted. Please see latest push.

tkilias (approved):

PR #225, opened 2024-11-22
Refactoring/#221 fixed mypy warnings

### LINKED ISSUE #221: Fix mypy warnings
https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/
https://mypy.readthedocs.io/en/stable/running_mypy.html

### LINKED ISSUE #222: Make BackgroundListenerThread.poller private
https://github.com/exasol/advanced-analytics-framework/blob/fcbf1bf32a3868fe3d85063d4140c31b1f47a623/exasol/analytics/udf/communication/peer_communicator/background_listener_thread.py#L141

### LINKED ISSUE #223: Typo catched -> caught
https://github.com/exasol/advanced-analytics-framework/blob/fcbf1bf32a3868fe3d85063d4140c31b1f47a623/exasol/analytics/query_handler/udf/runner/udf.py#L113C32-L113C39

ckunki:
Closes #221 
Closes #222
Closes #223

### DIFF
--- a/exasol/analytics/query_handler/context/proxy/bucketfs_location_proxy.py
+++ b/exasol/analytics/query_handler/context/proxy/bucketfs_location_proxy.py
@@ -1,6 +1,6 @@
 import logging
 
-import exasol.bucketfs as bfs
+import exasol.bucketfs as bfs # type: ignore[import-untyped]
 
 from exasol.analytics.query_handler.context.proxy.object_proxy import ObjectProxy
 
--- a/exasol/analytics/query_handler/deployment/lua_script_bundle.py
+++ b/exasol/analytics/query_handler/deployment/lua_script_bundle.py
@@ -26,20 +26,19 @@ class LuaScriptBundle:
 
     def bundle_lua_scripts(self, output_buffer: IO):
         with tempfile.TemporaryDirectory() as tmp_dir:
-            tmp_dir = Path(tmp_dir)
-            self.copy_lua_source_files(tmp_dir)
-            self.run_lua_amlg(tmp_dir, output_buffer)
+            path = Path(tmp_dir)
+            self.copy_lua_source_files(path)
+            self.run_lua_amlg(path, output_buffer)
 
     def copy_lua_source_files(self, tmp_dir: Path):
         for src in self.lua_source_files + [self.lua_main_file]:
             dst = tmp_dir / src.name
             logger.debug(f"Copy {src} to {tmp_dir}")
-            shutil.copy(src, dst)
+            shutil.copy(str(src), dst)
 
     def run_lua_amlg(self, tmp_dir: Path, output_buffer: IO):
         output_file = tmp_dir / f"bundle_{time.time()}.lua"
         bash_command = "amalg.lua -o {out_path} -s {main_file} {modules}".format(
-            tmp_dir=tmp_dir,
             out_path=output_file,
             main_file=self.lua_main_file.name,
             modules=" ".join(self.lua_modules),
--- a/exasol/analytics/query_handler/graph/result.py
+++ b/exasol/analytics/query_handler/graph/result.py
@@ -84,9 +84,9 @@ class _Meta(type):
             return _new(cls, result_type)
 
         result_type.__new__ = _configured_new
-        result_type.__init__ = _init
-        result_type.__setattr__ = _setattr
-        result_type.__delattr__ = _delattr
+        result_type.__init__ = _init # type: ignore
+        result_type.__setattr__ = _setattr # type: ignore
+        result_type.__delattr__ = _delattr # type: ignore
         return result_type
 
 
--- a/exasol/analytics/query_handler/graph/result.py
+++ b/exasol/analytics/query_handler/graph/result.py
@@ -79,11 +79,11 @@ class _Meta(type):
         """
         result_type = type(name, bases, attrs)
 
-        def _configured_new(cls: Type[cls]):
+        def _configured_new(cls: Type[_T]):
             """This function is called for subclasses of classes that declare _Meta as their metaclass."""
             return _new(cls, result_type)
 
-        result_type.__new__ = _configured_new
+        result_type.__new__ = _configured_new # type: ignore
         result_type.__init__ = _init # type: ignore
         result_type.__setattr__ = _setattr # type: ignore
         result_type.__delattr__ = _delattr # type: ignore
--- a/exasol/analytics/query_handler/graph/stage/sql/execution/query_handler_state.py
+++ b/exasol/analytics/query_handler/graph/stage/sql/execution/query_handler_state.py
@@ -21,6 +21,7 @@ from exasol.analytics.query_handler.graph.stage.sql.sql_stage_query_handler impo
 )
 from exasol.analytics.query_handler.query_handler import QueryHandler
 from exasol.analytics.query_handler.result import Continue, Finish
+from exasol.analytics.utils.errors import UninitializedAttributeError
 
 
 class ResultHandlerReturnValue(enum.Enum):
@@ -58,20 +59,33 @@ class SQLStageGraphExecutionQueryHandlerState:
         self._current_query_handler_context: Optional[ScopeQueryHandlerContext] = None
         self._create_current_query_handler()
 
-    def _check_is_valid(self):
-        if self._current_query_handler is None:
-            raise RuntimeError("No current query handler set.")
-
     def get_current_query_handler(
-        self,
+            self
     ) -> QueryHandler[List[SQLStageInputOutput], SQLStageInputOutput]:
-        self._check_is_valid()
-        return self._current_query_handler
+        value = self._current_query_handler
+        if value is None:
+            raise RuntimeError("No current query handler set.")
+        return value
+
+    @property
+    def current_query_handler_context(self) -> ScopeQueryHandlerContext:
+        value = self._current_query_handler_context
+        if value is None:
+            raise UninitializedAttributeError("current query handler context is None.")
+        return value
+
+    @property
+    def current_stage(self) -> SQLStage:
+        value = self._current_stage
+        if value is None:
+            raise UninitializedAttributeError("current stage is None.")
+        return value
 
     def handle_result(
         self, result: Union[Continue, Finish[SQLStageInputOutput]]
     ) -> ResultHandlerReturnValue:
-        self._check_is_valid()
+        # _check_is_valid()
+        self.get_current_query_handler()
         if isinstance(result, Finish):
             return self._handle_finished_result(result)
         elif isinstance(result, Continue):
@@ -90,7 +104,7 @@ class SQLStageGraphExecutionQueryHandlerState:
         return self._try_to_move_to_next_stage()
 
     def _try_to_move_to_next_stage(self) -> ResultHandlerReturnValue:
-        self._current_query_handler_context.release()
+        self.current_query_handler_context.release()
         if self._is_not_last_stage():
             self._move_to_next_stage()
             return ResultHandlerReturnValue.CONTINUE_PROCESSING
@@ -123,12 +137,12 @@ class SQLStageGraphExecutionQueryHandlerState:
             result_bucketfs_location=result_bucketfs_location,
             sql_stage_inputs=stage_inputs,
         )
-        self._current_query_handler = self._current_stage.create_train_query_handler(
+        self._current_query_handler = self.current_stage.create_train_query_handler(
             stage_input, self._current_query_handler_context
         )
 
     def _add_result_to_successors(self, result: SQLStageInputOutput):
-        successors = self._sql_stage_graph.successors(self._current_stage)
+        successors = self._sql_stage_graph.successors(self.current_stage)
         if len(successors) == 0:
             raise RuntimeError("Programming error")
         self._add_result_to_inputs_of_successors(result, successors)
@@ -146,7 +160,7 @@ class SQLStageGraphExecutionQueryHandlerState:
         object_proxies = find_object_proxies(result)
         for object_proxy in object_proxies:
             if object_proxy not in self._reference_counting_bag:
-                self._current_query_handler_context.transfer_object_to(
+                self.current_query_handler_context.transfer_object_to(
                     object_proxy, self._query_handler_context
                 )
             for _ in successors:
@@ -160,7 +174,7 @@ class SQLStageGraphExecutionQueryHandlerState:
                     object_proxy
                 )
             else:
-                self._current_query_handler_context.transfer_object_to(
+                self.current_query_handler_context.transfer_object_to(
                     object_proxy, self._query_handler_context
                 )
 
--- a/exasol/analytics/query_handler/python_query_handler_runner.py
+++ b/exasol/analytics/query_handler/python_query_handler_runner.py
@@ -17,6 +17,7 @@ from exasol.analytics.query_handler.query_handler import QueryHandler
 from exasol.analytics.query_handler.result import Continue, Finish
 from exasol.analytics.query_handler.udf.runner.state import QueryHandlerRunnerState
 from exasol.analytics.sql_executor.interface import SQLExecutor
+from exasol.analytics.utils.errors import UninitializedAttributeError
 
 LOGGER = logging.getLogger(__file__)
 
@@ -116,6 +117,10 @@ class PythonQueryHandlerRunner(Generic[ParameterType, ResultType]):
     def _wrap_return_query(
         self, input_query: SelectQueryWithColumnDefinition
     ) -> Tuple[str, str]:
+        if self._state.input_query_query_handler_context is None:
+            raise UninitializedAttributeError(
+                "Attribute _state.input_query_query_handler_context is None"
+            )
         temporary_view_name = (
             self._state.input_query_query_handler_context.get_temporary_view_name()
         )
--- a/exasol/analytics/query_handler/python_query_handler_runner.py
+++ b/exasol/analytics/query_handler/python_query_handler_runner.py
@@ -72,8 +72,8 @@ class PythonQueryHandlerRunner(Generic[ParameterType, ResultType]):
         self._cleanup_query_handler_context()
         self._execute_queries(result.query_list)
         input_query_result = self._run_input_query(result)
-        result = self._state.query_handler.handle_query_result(input_query_result)
-        return result
+        _result = self._state.query_handler.handle_query_result(input_query_result)
+        return _result
 
     def _run_input_query(self, result: Continue) -> PythonQueryResult:
         input_query_view, input_query = self._wrap_return_query(result.input_query)
--- a/exasol/analytics/query_handler/query/result/udf_query_result.py
+++ b/exasol/analytics/query_handler/query/result/udf_query_result.py
@@ -51,7 +51,7 @@ class UDFQueryResult(QueryResult):
 
     def fetch_as_dataframe(
         self, num_rows: Union[str, int], start_col: int = 0
-    ) -> "pandas.DataFrame":
+    ) -> "pandas.DataFrame": # type: ignore
         df = self._ctx.get_dataframe(num_rows, start_col=self._start_col)
         self._initialized = True
         if df is None:
--- a/exasol/analytics/query_handler/udf/runner/udf.py
+++ b/exasol/analytics/query_handler/udf/runner/udf.py
@@ -8,13 +8,15 @@ from enum import Enum, auto
 from io import BytesIO
 from typing import Any, List, Optional, Tuple
 
-import exasol.bucketfs as bfs
-import joblib
+import exasol.bucketfs as bfs # type: ignore[import-untyped]
+import joblib # type: ignore[import-untyped]
 
 from exasol.analytics.query_handler.context.scope import ScopeQueryHandlerContext
+from exasol.analytics.query_handler.udf.interface import UDFQueryHandlerFactory
 from exasol.analytics.query_handler.context.top_level_query_handler_context import (
     TopLevelQueryHandlerContext,
 )
+from exasol.analytics.query_handler.query.interface import Query
 from exasol.analytics.query_handler.query.result.udf_query_result import UDFQueryResult
 from exasol.analytics.query_handler.query.select import SelectQueryWithColumnDefinition
 from exasol.analytics.query_handler.result import Continue, Finish, Result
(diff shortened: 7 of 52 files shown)


exasol/analytics/udf/communication/peer_communicator/background_listener_thread.py
```diff
@@ -411,18 +434,24 @@ def _create_register_peer_connection(self, message: messages.RegisterPeer):
...
-        self._register_peer_connection = RegisterPeerConnection(
+        return RegisterPeerConnection(
             predecessor=message.source,
             predecessor_send_socket_factory=predecessor_send_socket_factory,
             successor=message.peer,
             successor_send_socket_factory=successor_send_socket_factory,
             my_connection_info=self._my_connection_info,
         )
 
+    @property
```
ckunki, line 445: Property method `register_peer_connection` checks the private attribute `_register_peer_connection` and raises an exception when it is `None` to avoid code duplication.

Questions: 
* Q1) The names of the private attribute and the property differ on in the prefix `_` and potentially could be mixed up. Should we use different names that can easier be distinguished? 
* Q2) The property is public. Should it be private?

exasol/analytics/query_handler/graph/stage/sql/execution/query_handler_state.py
```diff
@@ -58,20 +59,35 @@ def __init__(
...
+    def current_query_handler_context(self) -> ScopeQueryHandlerContext:
+        value = self._current_query_handler_context
+        if value is None:
+            raise UninitializedAttributeError(
+                "Current query handler context is undefined."
+            )
+        return value
+
+    @property
+    def current_stage(self) -> SQLStage:
```
ckunki, line 80: Property method `current_stage` checks the private attribute `_current_stage` and raises an exception when it is `None` to avoid code duplication.

Questions: 
* Q1) The names of the private attribute and the property differ on in the prefix `_` and potentially could be mixed up. Should we use different names that can easier be distinguished? 
* Q2) The property is public. Should it be private?

exasol/analytics/schema/column_name_builder.py
```diff
@@ -37,6 +39,8 @@ def with_table_like_name(
         return self
 
     def build(self) -> ColumnName:
+        if self._name is None:
+            raise TypeCheckError("Name must not be None.")
```
ckunki, line 43: Is it OK to raise a `TypeCheckError` here (same as typeguard does)?
Raising a different Error such as `ValueError` would require to update the unit tests.

exasol/analytics/udf/communication/broadcast_operation.py
```diff
@@ -54,8 +55,15 @@ def _receive_from_localhost_leader(self) -> bytes:
         self._check_sequence_number(specific_message_obj=specific_message_obj)
         return frames[1].to_bytes()
 
+    @property
```
ckunki, line 58: Property method `multi_node_communicator` checks the private attribute `_multi_node_communicator` and raises an exception when it is `None` to avoid code duplication.

Questions: 
* Q1) The names of the private attribute and the property differ on in the prefix `_` and potentially could be mixed up. Should we use different names that can easier be distinguished? 
* Q2) The property is public. Should it be private?

exasol/analytics/udf/communication/gather_operation.py
```diff
@@ -94,18 +109,25 @@ def _send_local_leader_message_to_multi_node_leader(self):
             local_position=local_position, value_frame=value_frame
         )
 
+    @property
```
ckunki, line 112: Property method `multi_node_communicator` checks the private attribute `_multi_node_communicator` and raises an exception when it is `None` to avoid code duplication.

Questions: 
* Q1) The names of the private attribute and the property differ on in the prefix `_` and potentially could be mixed up. Should we use different names that can easier be distinguished? 
* Q2) The property is public. Should it be private?

exasol/analytics/query_handler/context/proxy/bucketfs_location_proxy.py
```diff
@@ -1,6 +1,6 @@
 import logging
 
-import exasol.bucketfs as bfs
+import exasol.bucketfs as bfs  # type: ignore[import-untyped]
```
tkilias, line 3: I think, we can do this in the pyproject.toml for all imports, but we probably should add py.typed to bucketfs-python

exasol/analytics/query_handler/graph/result.py
```diff
@@ -79,14 +79,14 @@ def __new__(cls, name: str, bases: Tuple[Type, ...], attrs: Any):
...
-        def _configured_new(cls: Type[cls]):
+        def _configured_new(cls: Type[_T]):
             """This function is called for subclasses of classes that declare _Meta as their metaclass."""
             return _new(cls, result_type)
 
-        result_type.__new__ = _configured_new
-        result_type.__init__ = _init
-        result_type.__setattr__ = _setattr
-        result_type.__delattr__ = _delattr
+        result_type.__new__ = _configured_new  # type: ignore
```
tkilias, line 86: yeah, here we probably have no other way.

exasol/analytics/query_handler/graph/stage/sql/execution/query_handler_state.py
```diff
@@ -58,20 +59,35 @@ def __init__(
...
         self,
     ) -> QueryHandler[List[SQLStageInputOutput], SQLStageInputOutput]:
-        self._check_is_valid()
-        return self._current_query_handler
+        value = self._current_query_handler
+        if value is None:
+            raise RuntimeError("No current query handler set.")
+        return value
+
+    @property
```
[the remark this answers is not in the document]
tkilias: If the original is private, the new one needs to be private, too. This means it needs another name. Good question, what the name should be. The same is true for current_stage property.  Thinking about a prefix like _checked_*.

exasol/analytics/query_handler/udf/runner/udf.py
```diff
@@ -67,26 +75,38 @@ class QueryHandlerStatus(Enum):
...
+        self._parameter = None
+
+    @property
+    def bucketfs_location(self) -> bfs.path.PathLike:
+        if self._bucketfs_location is None:
+            raise UninitializedAttributeError("BucketFS location is undefined.")
+        return self._bucketfs_location
+
+    @property
+    def parameter(self) -> UDFParameter:
```
tkilias, line 102: these should be private, too

exasol/analytics/query_handler/udf/runner/udf.py
```diff
@@ -67,26 +75,38 @@ class QueryHandlerStatus(Enum):
...
 
     def __init__(self, exa):
         self.exa = exa
-        self.bucketfs_location: Optional[bfs.path.PathLike] = None
-        self.parameter: Optional[UDFParameter] = None
+        self._bucketfs_location = None
+        self._parameter = None
+
+    @property
+    def bucketfs_location(self) -> bfs.path.PathLike:
```
tkilias, line 96: these should be private, too

exasol/analytics/query_handler/udf/runner/udf.py
```diff
@@ -187,10 +211,10 @@ def release_and_create_query_handler_context_if_input_query(
             current_state.top_level_query_handler_context.get_child_query_handler_context()
         )
 
-    def _get_parameter(self, ctx):
+    def _get_parameter(self, ctx) -> UDFParameter:
         iter_num = ctx[0]
```
tkilias, line 215: in another PR, we probably should replace the numbers with named constants

exasol/analytics/query_handler/udf/runner/udf.py
```diff
@@ -260,9 +297,11 @@ def _remove_previous_state(self) -> None:
         self._state_file_bucketfs_location().rm()
 
     def _create_udf_query_result(
-        self, ctx, query_columns: List[Column]
+        self, ctx, query_columns: Optional[List[Column]]
     ) -> UDFQueryResult:
         colum_start_ix = 8 if self.parameter.iter_num == 0 else 4
+        if query_columns is None:
```
tkilias, line 303: questionable, needs more investigation

exasol/analytics/query_handler/python_query_handler_runner.py
```diff
@@ -116,24 +117,28 @@ def _release_and_create_query_handler_context_of_input_query(self):
     def _wrap_return_query(
         self, input_query: SelectQueryWithColumnDefinition
     ) -> Tuple[str, str]:
+        if self._state.input_query_query_handler_context is None:
```
tkilias, line 120: needs investigation

exasol/analytics/schema/column_name_builder.py
tkilias: it is not  a typeguard typecheck, so I would create my own TypeCheckError. However, we can do this also in another PR.

exasol/analytics/query_handler/udf/runner/udf.py
ckunki: OK

exasol/analytics/query_handler/udf/runner/udf.py
ckunki: OK

exasol/analytics/udf/communication/broadcast_operation.py
ckunki: See last push

exasol/analytics/query_handler/graph/stage/sql/execution/query_handler_state.py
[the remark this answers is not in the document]
ckunki: See last push

exasol/analytics/udf/communication/peer_communicator/background_listener_thread.py
ckunki: See last push

exasol/analytics/udf/communication/gather_operation.py
ckunki: See last push

exasol/analytics/query_handler/context/proxy/bucketfs_location_proxy.py
ckunki: See last push

exasol/analytics/query_handler/udf/runner/udf.py
ckunki: See #228

exasol/analytics/query_handler/deployment/lua_script_bundle.py
```diff
@@ -26,20 +26,19 @@ def __init__(
 
     def bundle_lua_scripts(self, output_buffer: IO):
         with tempfile.TemporaryDirectory() as tmp_dir:
-            tmp_dir = Path(tmp_dir)
-            self.copy_lua_source_files(tmp_dir)
-            self.run_lua_amlg(tmp_dir, output_buffer)
+            path = Path(tmp_dir)
+            self.copy_lua_source_files(path)
+            self.run_lua_amlg(path, output_buffer)
```
tomuben, line 31:
```suggestion
            tmp_path = Path(tmp_dir)
            self.copy_lua_source_files(tmp_path)
            self.run_lua_amlg(tmp_path, output_buffer)
```

exasol/analytics/query_handler/query/result/udf_query_result.py
```diff
@@ -51,7 +51,7 @@ def rowcount(self) -> int:
 
     def fetch_as_dataframe(
         self, num_rows: Union[str, int], start_col: int = 0
-    ) -> "pandas.DataFrame":
+    ) -> "pandas.DataFrame":  # type: ignore
```
tomuben, line 54: Would 
```suggestion
    ) -> Optional[pandas.DataFrame]: 
```
work here?

exasol/analytics/query_handler/context/proxy/bucketfs_location_proxy.py
```diff
@@ -1,4 +1,4 @@
-import logging
+import logging # xtype: ignore[import-untyped]
```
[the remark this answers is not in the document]
ckunki: Sorry, this was a left-overwhile experimenting with mypy settings in `pyproject.toml`.
Will be removed in next push.

exasol/analytics/query_handler/query/result/udf_query_result.py
ckunki: Yes, but requires importing pandas. I was not sure whether the string syntax was intentional, here?

exasol/analytics/query_handler/udf/runner/udf.py
```diff
@@ -67,29 +75,41 @@ class QueryHandlerStatus(Enum):
...
 
-@dataclasses.dataclass
+@dataclass
 class UDFResult:
     input_query_view: Optional[str] = None
     input_query: Optional[str] = None
-    final_result = {}
-    query_list = []
-    cleanup_query_list = []
+    final_result: str = "{}"
```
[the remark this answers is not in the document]
ckunki: Good point. Yes, looks like a functional change. On the other hand, after fixing the types a dict did not work anymore, a string did not break the tests and [in one place there is a string assigned](https://github.com/exasol/advanced-analytics-framework/blob/main/exasol/analytics/query_handler/udf/runner/udf.py#L127) to `final_result`, anyway. 

So, in the end, I am not sure about the correct solution.

I was not successful when declaring this attribute as a `Union[str, Dict]`.
Alternatively, we could wrap the stacktrace into a dict?
`final_result = { "stacktrace": stacktrace }`

exasol/analytics/query_handler/query/result/udf_query_result.py
ckunki: Importing pandas can take several seconds. 
See next push, replacing this with an explicit forward reference.

exasol/analytics/query_handler/udf/runner/udf.py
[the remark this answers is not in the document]
ckunki: See also the following code that ensures data type `str` to be used: 

https://github.com/exasol/advanced-analytics-framework/blob/9e3c4da41cb94b8a79679f26926bdb928c780903/exasol/analytics/query_handler/udf/runner/udf.py#L357

exasol/analytics/query_handler/query/result/udf_query_result.py
ckunki: I still get errors. See also https://github.com/pydantic/pydantic/issues/234
As far as I can see, either mypy complains or pytest fails.
I don't see a way to use `Optional` together with a type, that is not imported, yet.

exasol/analytics/udf/communication/peer_communicator/background_listener_interface.py
```diff
@@ -56,79 +76,75 @@ def __init__(
...
+            generic = deserialize_message(received, messages.Message)
+            message = generic.__root__
+            if not isinstance(message, messages.MyConnectionInfo):
+                raise UnexpectedMessageError(
+                    f"Unexpected message of type {type(message)}."
+                )
+            return message.my_connection_info
         except Exception as e:
-            self._logger.exception("Exception", raw_message=message)
+            self._logger.exception("Exception", raw_message=received)
```
tomuben, line 115: What if the exception happened in self._out_control.socket.receive()?
Then variable `received` is uninitialized, and thus the log would be useless.

exasol/analytics/udf/communication/peer_communicator/background_listener_thread.py
```diff
@@ -96,6 +98,50 @@ def create_background_peer_state_builder() -> BackgroundPeerStateBuilder:
...
+class RuntimeSockets:
+    in_control: Socket
+    out_control: Socket
+    listen: Socket
+    listen_port: int
+    poller: Poller
+
+    @classmethod
+    def create(
+        cls,
```
tomuben, line 111: Maybe define a generics for `cls`, see https://docs.python.org/3/library/typing.html#generics

exasol/analytics/udf/communication/socket_factory/fault_injection.py
```diff
@@ -64,7 +64,7 @@ def receive(self) -> bytes:
     def receive_multipart(self) -> List[abstract.Frame]:
         message = self._internal_socket.receive_multipart()
         converted_message = [Frame(frame) for frame in message]
-        return converted_message
+        return converted_message  # type: ignore
```
tomuben, line 67: Why do you need the type ignore comment here?

exasol/analytics/udf/communication/socket_factory/zmq_wrapper.py
```diff
@@ -149,7 +149,7 @@ def poll(
             self._sockets_map[zmq_socket]: _bitmask_to_flags(bitmask)
             for zmq_socket, bitmask in poll_result.items()
         }
-        return result
+        return result  # type: ignore
```
tomuben, line 152: Why do you need the type ignore here?

exasol/analytics/udf/utils/context_wrapper.py
```diff
@@ -28,7 +25,9 @@ def __getattr__(self, name):
 
     def get_dataframe(
         self, num_rows: Union[str, int], start_col: int = 0
-    ) -> Optional[pd.DataFrame]:
+    ) -> Optional[ForwardRef("pandas.DataFrame")]:  # type: ignore[valid-type]
```
tomuben, line 28: Can the "type: ignore" message now be removed?

exasol/analytics/udf/utils/iterators.py
```diff
@@ -1,6 +1,6 @@
 from typing import Any, Callable
 
-import pandas as pd
+import pandas as pd  # type: ignore
```
tomuben, line 3: Can be removed?

exasol/analytics/utils/errors.py
```diff
@@ -0,0 +1,12 @@
+class UninitializedAttributeError(Exception):
+    """
+    A method of a class accesses an attribute that has not been
+    initialized dynamically before, e.g. in method run().
+    """
+
+
+class IllegalParametersError(Exception):
+    """
+    Calling a function, method or iniinitializing a class with an illegal
```
tomuben, line 10:
```suggestion
    Indicating an error  when called a function, method or initialization a class with an illegal
```

tomuben (changes requested):
Don't forget to activate the type checker in the CI !!! Otherwise all the work is in vain ;)

exasol/analytics/udf/communication/messages.py
```diff
@@ -13,7 +13,7 @@ class BaseMessage(BaseModel, frozen=True):
 class RegisterPeer(BaseMessage, frozen=True):
     message_type: Literal["RegisterPeer"] = "RegisterPeer"
     peer: Peer
-    source: Optional["Peer"]
+    source: Optional["Peer"] = None
```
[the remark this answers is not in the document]
ckunki: Please see the results of my investigation in [discussion_r1860537415](https://github.com/exasol/advanced-analytics-framework/pull/225#discussion_r1860537415) .

TL;DR: not required and not recommended. Using string "Peer" is also shorter.
Only `from __future__ import annotations` could be an option that might allow to drop the quotes.

exasol/analytics/udf/communication/peer_communicator/background_listener_interface.py
ckunki: The original implementation used variable `message` which is 2 lines further down and hence with even more risk to be uninitialized. 
The logger is a structured logger (I can give additional explanation). 
Hence, a potentially uninitialized variable is acceptable from my perspective. 
We also added a `raise` statement after logging, which was missing before. 

All in all, I think the implementation is significantly improved and hopefully fine. 
But please only agree, if I could convince you.

exasol/analytics/udf/communication/socket_factory/fault_injection.py
ckunki: I got 
* "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
* Consider using "Sequence" instead, which is covariant
* Perhaps you need a type annotation for "converted_message"? Suggestion: "list[abstract.Frame]"

But adding this type annotation seems to fix the problem.

ckunki: See next push.

exasol/analytics/udf/communication/socket_factory/zmq_wrapper.py
ckunki: indeed - not needed. Thanks. 
See next push removing the ignore comment.

exasol/analytics/udf/utils/context_wrapper.py
ckunki: Currently, I don't see any perfect solution and hence propose to keep the comment.

Please see the results of my investigation in [discussion_r1860537415](https://github.com/exasol/advanced-analytics-framework/pull/225#discussion_r1860537415) .

ckunki:
> Don't forget to activate the type checker in the CI !!! Otherwise all the work is in vain ;)

I assume the type checker was activated all the time in [checks.yml](https://github.com/exasol/advanced-analytics-framework/blob/main/.github/workflows/checks.yml#L62):        
```
run: poetry run nox -s type-check
```

but mypy was told to ignore all package of the project in [pyproject.toml](https://github.com/exasol/advanced-analytics-framework/blob/main/pyproject.toml#60)

exasol/analytics/query_handler/udf/runner/udf.py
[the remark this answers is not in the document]
ckunki: See #229

exasol/analytics/udf/communication/peer_communicator/background_listener_thread.py
ckunki: This is a class method, compare to `self`.

tomuben (approved):

exasol/analytics/query_handler/udf/runner/udf.py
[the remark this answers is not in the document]
tkilias: Fixing of the missing Generic Types for QueryHandlerRunnerState will happen in https://github.com/exasol/advanced-analytics-framework/issues/230

tkilias (approved):

pyproject.toml
```diff
@@ -26,9 +26,7 @@ importlib-resources = "^6.4.0"
 exasol-bucketfs = ">=0.6.0,<1.0.0"
 click = "^8.0.4"
 pyexasol = ">=0.25.0,<1.0.0"
-pandas = "^1.1.0"
-# request from Mikhail:
-# pandas = "^2.2.3"
+pandas = "^2.2.3"
 pydantic = "^1.10.2"
```
tkilias, line 30: We should upgrade pydantic to version 2 it might otherwise conflict with other projects

ckunki: See ticket
* #233

PR #242, opened 2024-12-20
Added some comments and docstrings

ahsimb:


exasol/analytics/query_handler/graph/stage/sql/dataset.py
```diff
@@ -13,6 +13,9 @@ class Dataset:
 
     The TableLike refers to a database table containing the actual data that
     can be used for instance in training or testing.
+
+    Q. A TableLike is basically a list of columns and a name. Why do we have
+    a separate list of columns here?
```
ckunki, line 18: AFAIK there was and is a separation of "ordinary" columns and "identifier columns" which may be handled specially in some cases? But the attribute `columns` could be superfluous in my understanding.

exasol/analytics/query_handler/graph/stage/sql/execution/input.py
```diff
@@ -10,6 +10,12 @@
 
 @dataclasses.dataclass(frozen=True, eq=True)
 class SQLStageGraphExecutionInput:
+    """
+    The class is an input for the graph query handler (not to be confused with
+    a node query handler provided by a user).
```
ckunki, line 15: I am not familiar with the term "node query handler". 
Additionally, I propose, to add comments about the graph query handler to its class definition rather than to the `SQLStageGraphExecutionInput`.

ahsimb: There are two query handlers - one for the graph, implemented here, and one (or many) for each node, provided by the user.

ahsimb: There a docstring for the `SQLStageGraphExecutionQueryHandler`, albeit a short one.

exasol/analytics/query_handler/graph/stage/sql/sql_stage.py
```diff
@@ -9,6 +9,12 @@
 
 
 class SQLStage(Stage):
+    """
+    This is a node of an ExecutionGraph.
+    Essentially, this is a node-level query handler factory. The query handler
+    itself is user-provided and so is this factory.
```
ckunki, line 15:
```suggestion
    itself is user-provided and so are the factories implementing this interface.
```

exasol/analytics/query_handler/graph/stage/sql/input_output.py
```diff
@@ -8,16 +8,21 @@
...
 class SQLStageInputOutput:
     """
-    A SQLStageInputOutput is used as input and output between the SQLStageQueryHandler.
-    It contains a dataset and dependencies. The dataset is used to represent train and test data.
-    The dependencies can be used to communicate any data to the subsequently stages.
-    For example, a dependency could be a table which the previous stage computed and
-    the subsequent one uses.
+    This is a type root for a class representing an input/output data for a customer
+    provided node-level query handler - SQLStageQueryHandler. The actual content of
+    the input/output is application specific.
```
ckunki, line 13:
```suggestion
    This is a type root for a class representing input/output data for a customer
    provided node-level query handler extending class SQLStageQueryHandler. The actual content of
    the input/output is application-specific.
```

exasol/analytics/query_handler/graph/stage/sql/execution/query_handler_state.py
```diff
@@ -98,6 +105,11 @@ def handle_result(
     def _handle_finished_result(
         self, result: Finish[SQLStageInputOutput]
     ) -> ResultHandlerReturnValue:
+        """
+        We are done with the current node.
+        If there are successors we need to add the final output from this node to the
```
ckunki, line 110:
```suggestion
        If there are successors, then we need to add the final output from this node to the
```

ckunki (changes requested):

exasol/analytics/query_handler/graph/stage/sql/dataset.py
ahsimb: Yeah, that's what I meant. The `identifier_columns` OK, I can understand. But just `columns` are inside the `table_like` and then also separately. Why?

PR #256, opened 2025-02-25
#255 Added Query Handler audit interface

### LINKED ISSUE #255: Add means to pass audit data in the Continue and Finish objects
This includes:
- A dataclass for the user audit data.
- This dataclass should be included into the Finish object.
- a new query type, e.g. `SelectAuditQuery` to distinguish between normal and audit queries. The query handler may then add some queries for auditing in its `Continue` return and wrap them in this class. The output of such a query should match the above dataclass.

ahsimb:
closes #255

### DIFF
--- a/exasol/analytics/query_handler/audit/query_handler_audit.py
+++ b/exasol/analytics/query_handler/audit/query_handler_audit.py
@@ -1,0 +1,17 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class QueryHandlerAudit:
+    """
+    Definition of the audit data provided by a QueryHandler (QH).
+
+    The data can be provided in two ways.
+    1. An instance of this class can be included in the `Finish` object returned
+       by the `QH.start` or `QH.handle_query_result` in its final iteration.
+    2. An AuditQuery can return results with columns matching the attributes of
+       this class.
+    """
+    configuration: str | None = None
+    information: str | None = None
+    error: str | None = None
--- a/exasol/analytics/query_handler/query/select.py
+++ b/exasol/analytics/query_handler/query/select.py
@@ -1,4 +1,3 @@
-from abc import abstractmethod
 from typing import List
 
 from exasol.analytics.query_handler.query.interface import Query
@@ -6,6 +5,9 @@ from exasol.analytics.schema import Column
 
 
 class SelectQuery(Query):
+    """
+    A query wrapper for a read-only query. Such query is not auditable.
+    """
 
     def __init__(self, query_string: str):
         self._query_string = query_string
@@ -13,6 +15,34 @@ class SelectQuery(Query):
     @property
     def query_string(self) -> str:
         return self._query_string
+
+
+class WriteQuery(SelectQuery):
+    """
+    A query wrapper for a query that changes data in a database table (e.g. INSERT or
+    UPDATE) or creates the table (e.g. CREATE TABLE). This type of query is auditable.
+    """
+    def __init__(self, query_string: str, affected_table: str, audit: bool = False):
+        super().__init__(query_string)
+        self._affected_table = affected_table
+        self._audit = audit
+
+    @property
+    def affected_table(self) -> str:
+        return self._affected_table
+
+    @property
+    def audit(self) -> bool:
+        return self._audit
+
+
+class AuditQuery(SelectQuery):
+    """
+    A query wrapper for a special read-only query that selects data for auditing.
+    The output columns of such a query should correspond to the definition of the
+    audit data provided by a QueryHandler.
+    """
+    pass
 
 
 class SelectQueryWithColumnDefinition(SelectQuery):
--- a/exasol/analytics/query_handler/result.py
+++ b/exasol/analytics/query_handler/result.py
@@ -7,6 +7,7 @@ from typing import (
 
 from exasol.analytics.query_handler.query.interface import Query
 from exasol.analytics.query_handler.query.select import SelectQueryWithColumnDefinition
+from exasol.analytics.query_handler.audit.query_handler_audit import QueryHandlerAudit
 
 
 @dataclass()
@@ -26,3 +27,4 @@ T = TypeVar("T")
 @dataclass()
 class Finish(Generic[T], Result):
     result: T
+    audit: QueryHandlerAudit | None = None
--- a/doc/changes/unreleased.md
+++ b/doc/changes/unreleased.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+## Features
+
+* #255: Added means for a Query Handler to control the audit pass explicit information for the audit.
+
 ## Refactorings
 
 * #224: Replaced references to old BucketFS API with new API


exasol/analytics/query_handler/query/select.py
```diff
@@ -15,6 +17,34 @@ def query_string(self) -> str:
...
+    @property
+    def audit(self) -> bool:
+        return self._audit
+
+
+class AuditQuery(SelectQuery):
+    """
+    A query wrapper for a special read-only query that selects data for auditing.
+    The output columns of such a query should correspond to the definition of the
+    audit data provided by a QueryHandler.
```
ckunki, line 43: I am not sure about, who actually defines the resp. structure. 
Is it an arbitrary query handler?
Can different query handlers have different audit columns?

exasol/analytics/query_handler/query/select.py
```diff
@@ -15,6 +17,34 @@ def query_string(self) -> str:
         return self._query_string
 
 
+class WriteQuery(SelectQuery):
+    """
+    A query wrapper for a query that changes data in a database table (e.g. INSERT or
+    UPDATE) or creates the table (e.g. CREATE TABLE). This type of query is auditable.
+    """
+    def __init__(self, query_string: str, affected_table: str, audit: bool = False):
```
tkilias, line 25: if we want to do it this genric , I would like to add something like a operation (CREATE TABLE, INSERT TABLE)

exasol/analytics/query_handler/query/select.py
ahsimb: It's the QueryHandlerAudit

exasol/analytics/query_handler/query/select.py
```diff
@@ -15,6 +17,34 @@ def query_string(self) -> str:
         return self._query_string
 
 
+class WriteQuery(SelectQuery):
```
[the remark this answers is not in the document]
ahsimb: Not that I feel strongly about this, but yes, we already have the query_string implementation.

exasol/analytics/query_handler/audit/query_handler_audit.py
```diff
@@ -0,0 +1,17 @@
...
+
+    The data can be provided in two ways.
+    1. An instance of this class can be included in the `Finish` object returned
+       by the `QH.start` or `QH.handle_query_result` in its final iteration.
+    2. An AuditQuery can return results with columns matching the attributes of
+       this class.
+    """
+    configuration: str | None = None
+    information: str | None = None
+    error: str | None = None
```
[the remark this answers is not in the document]
ahsimb: The information may contain some explanation of what we were doing and what we tried to achieve. The error is what we have ended up with. I guess in most cases the combination of configuration and error will be enough, but who knows? The level or type, in this case, is just "good" vs "not good", and this can be derived from the error being None or not None.

exasol/analytics/query_handler/query/select.py
ahsimb: Yes, sure, why not?

exasol/analytics/query_handler/query/select.py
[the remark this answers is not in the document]
tkilias: ok, I also now understand, why you want to rename it to QueryWrapper or something. I think, it is a bit of a semantic misunderstanding between us. With SelectQuery, I meant really SELECT statements. I would agree that we rename the SelectQuery (not sure to what exact, maybe CustomQuery, there are other queries that don't have a query_string in the constructor) and then redefine the SelectQuery as a child class of this new class CustomQuery.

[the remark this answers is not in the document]
tkilias: AuditQuery would then inherit from CustomQuery instead of SelectQuery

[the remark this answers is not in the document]
ahsimb: We have the `audit` parameter that turns the audit on/off. The name `AuditWriteQuery` suggests that the audit is always on.

exasol/analytics/query_handler/query/select.py
tkilias: In my opinion, yes, different QueryHandlers can have different audit columns. That's why I wrote, QueryHandlers should have a function to tell which additional audit columns it would provide. This way we can go the call stack down and collect all columns. However, there should be a fixed subset, that all provide. The main difficulty is how to collect the value over the call stack, because I would like to have, that the leaves don't need to know anything about the columns provided by the QueryHandler in the middle.  Current, idea is a kind of Matroska QueryHandlerAudit, where each step wrap the QueryHandlerAudit it received from its child into its own.

exasol/analytics/query_handler/query/select.py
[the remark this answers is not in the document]
ahsimb: Re: `CustomQuery`,  I am fine with the class structure. As for the name, it's hard for me to comment. What is the meaning of the abstract class `Query`? As always, there is no documentation.

exasol/analytics/query_handler/query/select.py
ahsimb: This is something more complicated than what I had in mind. My idea was that the audit fields provided by a Query Handler (all or most are optional) are defined in the QueryHandlerAudit dataclass. Then there are two ways for the QH to pass them to the Executor. 1. As an instance of this class. 2. As a query that once executed will return columns corresponding to the fields of the dataclass.

ahsimb: So while you are executing queries in the `query_list` you can see a query wrapped in the `AuditQuery` wrapper. Then you transform this query into f'INSERT INTO AuditTable {the_query}'

exasol/analytics/query_handler/query/select.py
```diff
@@ -24,3 +28,46 @@ def __init__(self, query_string: str, output_columns: List[Column]):
     @property
     def output_columns(self) -> List[Column]:
         return self._output_columns
+
+
+class AuditData:
+    """
+    This is a collection of data for auditing. The data represent one audit message.
+    The items in the dictionary correspond to the columns in the audit table.
+    Components at different levels in the call stack can add their own items here.
```
tkilias, line 37: Regarding this point, I thought a bit, and I got a question for us, should we make the dict append only, meaning that higher levels in the call stack can only add new entries to the dict, but can't change already set ones. My first impression is, that would reduce surprises. What do you think?

exasol/analytics/query_handler/query/select.py
```diff
@@ -24,3 +28,46 @@ def __init__(self, query_string: str, output_columns: List[Column]):
...
+    """
+    def __init__(self, audit_fields: dict[str, Any] | None = None):
+        self._audit_fields = audit_fields or {}
+
+    @property
+    def audit_fields(self) -> dict[str, Any]:
+        return self._audit_fields
+
+
+class AuditQuery(AuditData):
```
tkilias, line 47:
```suggestion
class AuditQuery(AuditData, Query):
```

this way, we can inject the AuditQuery without problems into the query_list of Continue. FYI, Query is an interface that requests to implement the property query_text, which would be in this case simply the forwarding of the query_text of the _query_with_columns

exasol/analytics/query_handler/result.py
```diff
@@ -26,3 +27,4 @@ class Continue(Result):
 @dataclass()
 class Finish(Generic[T], Result):
     result: T
+    audit_data: AuditData | None = None
```
tkilias, line 30: I think, we don't want to allow a write query here

```suggestion
    audit_data: AuditQuery | None = None
```

ckunki: `AuditData` is not a _query_ (neither write nor read), but just "data". 
Not sure how this relates to "write" or "read"?

exasol/analytics/query_handler/query/select.py
```diff
@@ -24,3 +28,46 @@ def __init__(self, query_string: str, output_columns: List[Column]):
...
+    def __init__(self, query_with_columns: SelectQueryWithColumnDefinition,
+                 audit_fields: dict[str, Any] | None = None):
+        super().__init__(audit_fields)
+        self._query_with_columns = query_with_columns
+
+    def query_with_columns(self) -> SelectQueryWithColumnDefinition:
+        return self._query_with_columns
+
+
+class WriteQuery(CustomQuery, AuditData):
```
ckunki, line 60: I am not sure whether inheritance is the best way to model the aspect of audit logging.
I cannot come up with a proposal that is clearly better at the moment, though.

On the other hand (at least in Java) I often needed to learn, that composition (pythonic: using an attribute) is better than inheritance.

Also, I see multiple dimensions:
- database operations select, create, update, delete which probably benefit from explicit classes like `CreateQuery` vs. `UpdateQuery`, for example
- the aspect _read_ or _write_ which clearly can be derived from the database operations select = read, vs. Create, Update, Delete = write
- the aspect of audit logging that obviously is in focus here
  - audit log queries in themselves for my understanding are read queries
  - creating entries in the audit log is probably planned only for write queries

And whenever there are multiple dimensions, the hierarchical concept of inheritance doesn't feel perfect to me.

I am sorry, if this might sound rather philosophical and as I cannot provide a summary or even a call for action in the end.

ckunki (commented):
See my comments.

exasol/analytics/query_handler/query/select.py
```diff
@@ -24,3 +28,46 @@ def __init__(self, query_string: str, output_columns: List[Column]):
...
+    """
+    A query wrapper for a query that changes data in a database table (e.g. INSERT or
+    UPDATE) or creates the table (e.g. CREATE TABLE). This type of query is auditable.
+    """
+    def __init__(self, query_string: str, audit: bool = False,
+                 audit_fields: dict[str, Any] | None = None):
```
tkilias, line 66: we need the table_name and operation type (CREATE TABLE, INSERT TABLE)

exasol/analytics/query_handler/query/select.py
tkilias: this also guarantees backwards compatibility with the Lua QueryHandler Runner which would run this query as is without doing logging. Because, audit logging, we only will add after Q2 to the Lua QueryHandler Runner.

tkilias:
```
def query_text():
    if self._select_query_with_columns is not None:
        return self._select_query_with_columns.query_text
   else:
       return "SELECT 1"
```

exasol/analytics/query_handler/query/select.py
```diff
@@ -24,3 +28,46 @@ def __init__(self, query_string: str, output_columns: List[Column]):
...
+    @property
+    def audit_fields(self) -> dict[str, Any]:
+        return self._audit_fields
+
+
+class AuditQuery(AuditData):
+    """
+    A query wrapper for a special read-only query that selects data for auditing.
+    """
+    def __init__(self, query_with_columns: SelectQueryWithColumnDefinition,
```
tkilias, line 51:
```suggestion
    def __init__(self, query_with_columns: Optional[SelectQueryWithColumnDefinition]=None,
```

exasol/analytics/query_handler/query/select.py
tkilias:
```
class InsertTableQuery:
  def init(..., table_name:str):
     audit_fields["table_name"] = table_nam
 @property
 def table_name()
```

exasol/analytics/query_handler/query/select.py
```diff
@@ -44,30 +50,41 @@ def audit_fields(self) -> dict[str, Any]:
...
     """
-    A query wrapper for a special read-only query that selects data for auditing.
+    A wrapper for a special read-only query that selects data for auditing. An object
+    of the class can also be used as an audit property bag, since it inherits from the
+    `AuditData` as well as `SelectQueryWithColumnDefinition`. The query is optional.
+    If provided the output columns should also be defined.
     """
-    def __init__(self, query_with_columns: SelectQueryWithColumnDefinition,
+    def __init__(self, query_string: str = '',
+                 output_columns: list[Column] | None = None,
```
tkilias, line 61: if output columns can get None here, we need to adapt the type hints in SelectQueryWithColumnDefinition, what speaks against it to have SelectQueryWithColumnDefinition as an optional attribute and inherit from Query instead and do a simple implementation for query_text property.

exasol/analytics/query_handler/query/select.py
```diff
@@ -44,30 +50,41 @@ def audit_fields(self) -> dict[str, Any]:
...
 class WriteQuery(CustomQuery, AuditData):
     """
-    A query wrapper for a query that changes data in a database table (e.g. INSERT or
-    UPDATE) or creates the table (e.g. CREATE TABLE). This type of query is auditable.
+    A wrapper for a query that changes data in a database table (e.g. INSERT or UPDATE)
+    or creates the table (e.g. CREATE TABLE). This type of query is auditable.
     """
-    def __init__(self, query_string: str, audit: bool = False,
-                 audit_fields: dict[str, Any] | None = None):
+    def __init__(self, query_string: str, affected_table: TableLikeName,
```
tkilias, line 77: if we do it this way, we probably should rename the class to WriteTableQuery and add another property operation, which can be of value INSERT, DELETE, CREATE, DROP, TRUNCATE

exasol/analytics/query_handler/query/select.py
ahsimb: This is the question of whether the AuditQuery IS Query or ENCAPSULATES a query. If it IS a Query, which should be assumed if the class inherits from `Query` then the way I wrote it is the correct way of doing this. The logic suffers a bit from the fact that it is now optional. Maybe setting the default value to "SELECT 1" will work better.

ahsimb: So that will be my proposal: Remove the validation, replace the empty string with "SELECT 1", so that we can always run the query from pyexasol - no checks needed. The columns should define the columns we collect from the results. If it's an empty list - fair game. I might have needed to execute a query for some reason but not collect anything for auditing. I can't think of any use case right now, but who knows?

exasol/analytics/query_handler/query/select.py
ahsimb: OK, if we also need to record the operation then we have two options - multiple classes or an extra parameter. I think they are equivalent. The latter is probably more compact.
The name `WriteTableQuery` kinda excludes the CREATE TABLE case. By `WriteQuery` I meant writing some to the database, as opposed to selecting the data and leaving the database unchanged.

exasol/analytics/query_handler/query/select.py
tkilias: I think, your proposal has one problem, and this makes it hard for us, to detect whether the query is intended or not to be logged.
if we have instead a separate select_with_column property, we know, if it is None, that we don't need it and regarding the implementation of the query_text property in the AuditQuery, we return either the query_text of te select_with_columns or SELECT 1

exasol/analytics/query_handler/query/select.py
ahsimb: What is db_object_type? Do you still want the INSERT, DELETE, CREATE, DROP, TRUNCATE?

tkilias: db_object_type would be TABLE, SCHEMA, ..., the INSERT, DELETE, CREATE would be the operation which also need. Having the db object name, type and the operation allows us to keep the Write/ModifyQuery very generic

tkilias: Good question, enumeration is probably saner, but maybe do, Enum | str, for the case of a consumer of the AAF has specific requirements

tkilias (changes requested):

tkilias (approved):

PR #275, opened 2025-03-24
Feature/#272 add class log span

### LINKED ISSUE #272: Audit Log: Add class LogSpan
```python
@dataclass
class LogSpan:
    parent_span: LogSpan | None
    name: str
    id: uuid

    def get_child_span(self, name: str) -> LogSpan:
        pass

class ModifyQuery:
    def init(db_operation_type: DbOperationType, parent_log_span: LogSpan):
        self.parent_log_span = parent_span
        self.log_span = (
            Span() if not parent_log_span
            else parent_log_span.get_child(db_operation_type)
        ) 

class AuditQuery:
    def init(log_span: LogSpan):
        self.log_span = log_span
```

See IntRef for additional details.

See https://docs.python.org/3.10/library/uuid.html

Examples: 
* `uuid.uuid4()` Generate a random UUID.

Generate a UUID based on the SHA-1 hash of a namespace identifier (which is a UUID) and a name (which is a string).
* `uuid.uuid5(uuid.NAMESPACE_DNS, "exasol.com")`
* `uuid.uuid5(uuid.NAMESPACE_URL, "https://exasol.com")`

Obtain UUID value as 128-bit integer: `UUID.int`

### LINKED ISSUE #273: Audit Log: Add RUN_ID
` POSIX_TIME(SYSTIMESTAMP(9)) * 1000`

ckunki:
Closes #272 
Closes #273


exasol/analytics/audit/audit.py
```diff
@@ -71,27 +91,33 @@ def augment(self, queries: Iterator[Query]) -> Iterator[str]:
                 )
 
     def _insert(self, query: AuditQuery) -> str:
```
ckunki, line 93: This code has been refactored and some parts have been moved to a central method `_insert_statement()`.

exasol/analytics/audit/audit.py
```diff
@@ -124,41 +150,64 @@ def _count_rows(self, query: ModifyQuery, event_name: str) -> str:
...
-            "LOG_SPAN_NAME": query.db_operation_type.name,
-            "EVENT_NAME": event_name,
-            "DB_OBJECT_TYPE": query.db_object_type.name,
-            "DB_OBJECT_SCHEMA": schema(db_obj),
-            "DB_OBJECT_NAME": db_obj.name,
-        }
-        other_table = query.db_object_name.fully_qualified
-        row_count = BaseAuditColumns.ROW_COUNT.name.name
+        return base_column_values(
+            {
```
ckunki, line 184: Replaced literal columns names by references to `BaseAuditColumns` and using function `base_column_values()` to convert into a dict with strings as keys.

ahsimb:
Does this PR closes 277?

exasol/analytics/audit/audit.py
```diff
@@ -23,12 +26,28 @@
 )
 
 
+def base_column_values(
```
ahsimb, line 29: It seems this function does two distinct things - changing the keys and adding another entry.

exasol/analytics/schema/column.py
```diff
@@ -50,3 +53,33 @@ def varchar_column(
...
         ),
     )
+
+
+def hashtype_column(
+    name: str,
+    bytes: Optional[int] = None,
+    bits: Optional[int] = None,
+) -> Column:
+    def size_and_unit() -> tuple[int, SizeUnit]:
```
ahsimb, line 63: What was the point in making it a function?

exasol/analytics/query_handler/query/select.py
```diff
@@ -104,13 +132,15 @@ def __init__(
         db_operation_type: DbOperationType,
         audit_fields: dict[str, Any] | None = None,
         audit: bool = False,
+        parent_log_span: LogSpan | None = None,
```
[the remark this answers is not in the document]
ckunki: This is due to the design, see IntRef of #272.
The idea (AFAIK) potentially is, that
* a `ModifyQuery` 
  * will create a span on its own, see method `_wrap()`, returning `_count_rows("Begin")`, `query`, `count_rows("End")`
  * is potentially child of a parent span
* while an `AuditQuery` 
  * is a leaf from the perspective of nested spans. 
  * the specified span could be called parent_span as well, but there is no child

exasol/analytics/schema/column.py
ckunki: * First of all, this is in line with the other functions, eg. `varchar_column()`, etc.
* Second: It does not only create a column type but also a column, with a name _and_ the type. 
* Third: The interface is not `bytes` and `bit` rather than `size` and `unit`
* Fourth: If you refer to the error checks, then, yes, I was tempted to move these to class `ColumnType`.

But moving the error checks to class `ColumnType`, would require more changes
* switch from `dataclass` to ordinary class 
* add `__init__`
* implement `__eq__` and `__repr__`

In the end, I was reluctant if I should add these changes to the current PR. 
What is your opinion?

exasol/analytics/audit/audit.py
ckunki: Yes, as described in the docstring. 
Would you prefer a different approach?

exasol/analytics/audit/audit.py
```diff
@@ -124,41 +150,64 @@ def _count_rows(self, query: ModifyQuery, event_name: str) -> str:
...
+            .add_scalar_functions(BaseAuditColumns.values | scalar_functions)
+            .add_constants(constants)
+            .add_references(references)
         )
         return (
             f"INSERT INTO {self.name.fully_qualified} (\n"
             f"  {insert_statement.columns}\n"
             ") SELECT\n"
-            f"  {insert_statement.values}"
+            f"  {insert_statement.values}{suffix}"
```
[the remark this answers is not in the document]
ckunki: Currently, there is no space. 
This enables to end the statement right after the `insert_statement.values` if there is no suffix.
Would you prefer to have a space character, even if the suffix is empty?

exasol/analytics/audit/audit.py
ahsimb: Personally, I don't take these things religiously, but the convention is that a function should do only one thing.

exasol/analytics/audit/audit.py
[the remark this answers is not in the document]
ahsimb: I prefer a note somewhere that the provided suffix  must have a space, if it is expected to have one.

exasol/analytics/schema/column.py
ahsimb: Not a big deal of course. You usually create a function within a function if you need to compute something repeatedly, e.g. in a cycle or list comprehension. If you want to improve code readability you would move a piece of code into a separate function. To me, an inner function called only once looks a bit odd.

exasol/analytics/audit/audit.py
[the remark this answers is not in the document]
ckunki: I added a docstring to method `_insert()`.

exasol/analytics/schema/column.py
ckunki: I wanted to separate the error handling and different values to be returned from the part creating the actual instance of `Column`. But looking at it, you maybe right. 
See next push removing the function.

exasol/analytics/query_handler/query/select.py
```diff
@@ -104,13 +135,15 @@ def __init__(
...
     ):
         CustomQuery.__init__(self, query_string)
         AuditData.__init__(self, audit_fields)
         self._db_object_type = db_object_type
         self._db_object_name = db_object_name
         self._db_operation_type = db_operation_type
         self._audit = audit
+        self._log_span = LogSpan(db_operation_type.name, parent=parent_log_span)
```
tkilias, line 146: is there a reason, you don't use parent_log_span.child(...)

exasol/analytics/query_handler/query/select.py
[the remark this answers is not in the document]
ckunki: Actually, @tkilias had some concerns as well. 
So I changed the implementation to create an explicit instance of `LogSpan` inside `ModifyQuery`.
I hope this also resolves your initial question?
Please have a look again.

exasol/analytics/query_handler/query/select.py
ckunki: OMG - you're right. I forgot about that. :confused:

ckunki: Now I know, why: `parent_log_span` is optional. 

The question is: Do we want to have a `LogSpan` for each `ModifyQuery`, even if `parent_log_span` is `None`?

tkilias (approved):

PR #279, opened 2025-03-27
 #274: Created AuditQueryHandler

### LINKED ISSUE #274: Audit Log: Create QueryHandler

ckunki:
Closes #274

### DIFF
--- a/exasol/analytics/audit/audit.py
+++ b/exasol/analytics/audit/audit.py
@@ -218,3 +218,12 @@ class AuditTable(Table):
             db_object_name=self.name,
             db_operation_type=DbOperationType.INSERT,
         )
+
+    @property
+    def create_query(self) -> ModifyQuery:
+        return ModifyQuery(
+            query_string=self.create_statement,
+            db_object_type=DbObjectType.TABLE,
+            db_object_name=self.name,
+            db_operation_type=DbOperationType.INSERT,
+        )
--- a/exasol/analytics/audit/audit_query_handler.py
+++ b/exasol/analytics/audit/audit_query_handler.py
@@ -1,0 +1,137 @@
+from __future__ import annotations
+from enum import Enum
+from typing import (
+    Callable,
+    Generic,
+    Iterator,
+    Optional,
+    TypeVar,
+    TypeAlias,
+    Union,
+    cast,
+)
+from dataclasses import dataclass
+from pydantic import BaseModel
+
+from exasol.analytics.query_handler.graph.stage.sql.sql_stage_graph import SQLStageGraph
+from exasol.analytics.query_handler.query.result.interface import QueryResult
+from exasol.analytics.query_handler.query_handler import QueryHandler
+from exasol.analytics.query_handler.result import (
+    Continue,
+    Finish,
+)
+from exasol.analytics.query_handler.query.interface import Query
+from exasol.analytics.query_handler.query.select import AuditQuery
+from exasol.analytics.audit.audit import AuditTable
+from exasol.analytics.query_handler.query.select import SelectQueryWithColumnDefinition
+from exasol.analytics.query_handler.context.scope import ScopeQueryHandlerContext
+
+class Phase(Enum):
+    MAIN = 2
+    RUN_FINAL_AUDIT_QUERY = 3
+    FINISHED = 4
+
+
+ResultType = TypeVar("ResultType")
+ParameterType = TypeVar("ParameterType")
+ContinueOrFinish: TypeAlias = Union[Continue, Finish[ResultType]]
+
+
+class IllegalMethodCallError(RuntimeError):
+    """
+    The application framework logic called a method that was not intended
+    to be called, e.g. ``IdentityQueryHandler.handle_query_result()``.
+    """
+
+
+@dataclass
+class Child(Generic[ResultType]):
+    context: ScopeQueryHandlerContext
+    query_handler: QueryHandler
+    result: ResultType | None
+
+
+class AuditQueryHandler(QueryHandler[ParameterType, ResultType]):
+    """
+    Use the specified factory (e.g. :class:`OrchestratorQueryHandler`) to
+    instantiate a query handler and augment the queries returned by it for
+    creating Audit Log entries.
+    """
+    def __init__(
+        self,
+        parameter: ParameterType,
+        context: ScopeQueryHandlerContext,
+        query_handler_factory: Callable[
+            [ParameterType, ScopeQueryHandlerContext],
+            QueryHandler[ParameterType, ResultType],
+        ],
+
+    ):
+        super().__init__(parameter, context)
+        self._phase = Phase.MAIN
+        self._audit_table: Optional[AuditTable] = None
+
+        child_context = context.get_child_query_handler_context()
+        query_handler = query_handler_factory(parameter, child_context)
+        self._child = Child[ResultType](child_context, query_handler, None)
+
+    def start(self) -> ContinueOrFinish:
+        action = self._child.query_handler.start()
+        return self._handle_action(action)
+
+    def handle_query_result(self, query_result: QueryResult) -> ContinueOrFinish:
+        if self._phase == Phase.MAIN:
+            action = self._child.query_handler.handle_query_result(query_result)
+            return self._handle_action(action)
+        elif self._phase == Phase.RUN_FINAL_AUDIT_QUERY:
+            self._phase = Phase.FINISHED
+            return Finish(result=self._child.result, audit_query=None)
+        else:
+            raise IllegalMethodCallError(
+                f"Method {type(self).__name__}.handle_query_result()"
+                f" has been called in phase {self._phase.name}."
+            )
+
+    def _handle_action(self, action: ContinueOrFinish) -> ContinueOrFinish:
+        if isinstance(action, Continue):
+            caction = cast(Continue, action)
+            return Continue(
+                query_list=self._augmented(caction.query_list),
+                input_query=caction.input_query,
+            )
+        elif isinstance(action, Finish):
+            self._child.context.release()
+            if action.audit_query:
+                self._child.result = action.result
+                self._phase = Phase.RUN_FINAL_AUDIT_QUERY
+                return self._final_continue(action.audit_query)
+            else:
+                self._phase = Phase.FINISHED
+                return action
+        else:
+            raise RuntimeError(f"Unknown action type {type(action)}")
+
+    def _final_continue(self, audit_query: AuditQuery) -> Continue:
+        """
+        Create Continue for excecuting the final AuditQuery.
+        """
+        input_query = SelectQueryWithColumnDefinition(
+            query_string="SELECT 1",
+            output_columns=[],
+        )
+        return Continue(
+            query_list=self._augmented([audit_query]),
+            input_query=input_query,
+        )
+
+    def _augmented(self, queries: list[Query]) -> list[Query]:
+        def generate() -> Iterator[Query]:
+            if self._audit_table is None:
+                self._audit_table = AuditTable(
+                    db_schema="<TBD>",
+                    table_name_prefix="<TBD>",
+                    additional_columns=[], # TBD
+                )
+                yield self._audit_table.create_query
+            yield from self._audit_table.augment(iter(queries))
+        return list(generate())
--- a/doc/changes/unreleased.md
+++ b/doc/changes/unreleased.md
@@ -6,6 +6,7 @@
 * #273: Enhanced AuditTable to set `RUN_ID`
 * #276: Added support for column type `HASHTYPE`
 * #277: Added `DELETE` to `DbOperationType`
+* #274: Created `AuditQueryHandler`
 
 ## Refactorings
 


exasol/analytics/audit/audit_query_handler.py
```diff
@@ -0,0 +1,143 @@
...
+                return action
+        else:
+            raise RuntimeError(f"Unknown action type {type(action)}")
+
+    def _final_continue(self, audit_query: AuditQuery) -> Continue:
+        """
+        Create Continue for excecuting the final AuditQuery.
+        """
+        input_query = SelectQueryWithColumnDefinition(
+            query_string="SELECT 1",
```
tkilias, line 124: cast decimal

exasol/analytics/audit/audit_query_handler.py
```diff
@@ -0,0 +1,143 @@
...
+            output_columns=[],
```
tkilias, line 125: return column decimal(1,0)

exasol/analytics/audit/audit_query_handler.py
```diff
@@ -0,0 +1,143 @@
...
+class AuditQueryHandler(QueryHandler[ParameterType, ResultType]):
+    """
+    Use the specified factory (e.g. :class:`OrchestratorQueryHandler`) to
+    instantiate a query handler and augment the queries returned by it for
+    creating Audit Log entries.
+    """
+
+    def __init__(
+        self,
+        parameter: ParameterType,
```
tkilias, line 68: class AuditQueryHandlerParameter[ParameterType]:

child_parameter:ParameterType
db_schema
table_prefix

tkilias: Theoretical the factory could contain the parameter which would make the parameter object easier

tkilias: Alternative, we don't have AuditQueryHandlerParameter, but we have functions as parameters, that extract db_schema and table_prefix from whatever Parameter we got

ckunki: See last push

exasol/analytics/audit/audit_query_handler.py
```diff
@@ -0,0 +1,143 @@
...
+        context: ScopeQueryHandlerContext,
```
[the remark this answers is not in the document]
ckunki: See last push

exasol/analytics/audit/audit_query_handler.py
ckunki: See last push

exasol/analytics/audit/audit.py
```diff
@@ -225,5 +225,5 @@ def create_query(self) -> ModifyQuery:
             query_string=self.create_statement,
             db_object_type=DbObjectType.TABLE,
             db_object_name=self.name,
-            db_operation_type=DbOperationType.INSERT,
+            db_operation_type=DbOperationType.CREATE_IF_NOT_EXISTS,
```
tkilias, line 228: if the audit table doesn't exist and you try to log the creation this will fail. We need to set audit ot false.

exasol/analytics/audit/audit_query_handler.py
```diff
@@ -71,20 +82,29 @@ def __init__(
             [ParameterType, ScopeQueryHandlerContext],
             QueryHandler[ParameterType, ResultType],
         ],
+        schema_getter: Callable[[ParameterType], str],
+        table_name_prefix_getter: Callable[[ParameterType], str],
+        additional_columns_provider: Callable[
```
tkilias, line 87: The additional columns are static and don't depend on the Parameter, they only depend on the caller of this constructor.

exasol/analytics/audit/audit_query_handler.py
```diff
@@ -121,8 +141,8 @@ def _final_continue(self, audit_query: AuditQuery) -> Continue:
         Create Continue for excecuting the final AuditQuery.
         """
         input_query = SelectQueryWithColumnDefinition(
-            query_string="SELECT 1",
-            output_columns=[],
+            query_string="SELECT 1 as DUMMY_COLUMN",
```
tkilias, line 144:
```suggestion
            query_string="SELECT CAST(1 as DECIMAL(1,0)) as DUMMY_COLUMN",
```

exasol/analytics/audit/audit_query_handler.py
```diff
@@ -121,8 +141,8 @@ def _final_continue(self, audit_query: AuditQuery) -> Continue:
...
+            output_columns=[decimal_column("DUMMY_COLUMN", precision=10)],
```
tkilias, line 145:
```suggestion
            output_columns=[decimal_column("DUMMY_COLUMN", precision=1, scale=0)],
```

exasol/analytics/schema/table.py
```diff
@@ -17,10 +17,8 @@ def __init__(self, name: TableName, columns: List[Column]):
...
         columns = ",\n  ".join(c.for_create for c in self.columns)
-        return cleandoc(
-            f"""
-            CREATE TABLE IF NOT EXISTS {self.name.fully_qualified} (
-              {columns}
-            )
-            """
+        return (
+            f"CREATE TABLE IF NOT EXISTS {self.name.fully_qualified} (\n"
+            f"  {columns}\n"
```
tkilias, line 22: the indentation only works for the first column

exasol/analytics/schema/table.py
```diff
@@ -17,10 +17,8 @@ def __init__(self, name: TableName, columns: List[Column]):
...
+            ")"
```
tkilias, line 23: I would suggest, to  concat the columns only with `,\n` and use [textwrap.indent](https://docs.python.org/3/library/textwrap.html#textwrap.indent)

tests/unit_tests/audit_log/test_audit_query_handler.py
```diff
@@ -0,0 +1,127 @@
...
+def test_constructor() -> None:
+    child = Mock()
+    testee = create_audit_query_handler(child)
+    audit_table = testee._audit_table
+    assert audit_table.name == AUDIT_TABLE_NAME
+    assert '"DDD" DECIMAL(9,0)' in [c.for_create for c in audit_table.columns]
+    assert child == testee._child.query_handler
+
+
+def test_start_finish():
```
tkilias, line 84: from here you should use the PythonQueryHandlerRunner with MockSQLExecutor, because it simulates the environment your QueryHandler runs in way better.

tests/unit_tests/audit_log/test_audit_query_handler.py
```diff
@@ -0,0 +1,127 @@
...
+        return [decimal_column("DDD", precision=9)]
+
+    parameter = MyParameterType(
+        db_schema="SSS",
+        table_name_prefix=AUDIT_TABLE_NAME_PREFIX,
+    )
+    return AuditQueryHandler(
+        parameter=parameter,
+        context=Mock(),
+        query_handler_factory=Mock(return_value=child),
```
tkilias, line 68: not sure, if using a Mock here is useful, because you don't use the Mock for testing. A lambda or internal function would do the job too

tests/unit_tests/audit_log/test_audit_query_handler.py
tkilias, line 84: From here, you should use the PythonQueryHandlerRunner, because it simulates the environment you are running better.

tkilias, line 84:
```suggestion
def test_start_finish_without_audit_query():
```

tests/unit_tests/audit_log/test_audit_query_handler.py
```diff
@@ -0,0 +1,127 @@
...
+
+def continue_action(query_list: list[Query]) -> Continue:
+    input_query = SelectQueryWithColumnDefinition(
+        query_string="SELECT 1 as DUMMY_COLUMN",
+        output_columns=[decimal_column("DUMMY_COLUMN", precision=10)],
+    )
+    return Continue(query_list=query_list, input_query=input_query)
+
+
+def test_start_continue_finish():
```
tkilias, line 99:
```suggestion
def test_start_continue_with_audit_query_finish_without_audit_query():
```

tests/unit_tests/audit_log/test_audit_query_handler.py
tkilias: Additional to these tests, you should also add tests with PythonQueryHandlerRunner, because it simulates better the restriction of the system, aka the column definition match for the return_query of the Continue.

tests/unit_tests/audit_log/test_audit_query_handler.py
```diff
@@ -0,0 +1,127 @@
...
+        expected_query(AUDIT_TABLE_NAME, DbOperationType.CREATE_IF_NOT_EXISTS),
+        expected_query(AUDIT_TABLE_NAME),
+        expected_query(query.db_object_name, query.db_operation_type),
+        expected_query(AUDIT_TABLE_NAME),
+    ]
+
+    testee = create_audit_query_handler(child)
+    action_1 = testee.start()
+    for expected, actual in zip(expected_queries, action_1.query_list):
+        assert_queries_match(expected, actual, QueryStringCriterion.STARTS_WITH)
```
tkilias, line 125: why don't you use `assert expected.startswith(actual)`

tests/utils/audit_table_utils.py
```diff
@@ -67,3 +73,40 @@ def create_insert_query(
...
+    }
+    return ModifyQuery(
+        query_string=query_strings[db_operation_type],
+        db_object_type=DbObjectType.TABLE,
+        db_object_name=table_name,
+        db_operation_type=db_operation_type,
+    )
+
+
+def assert_queries_match(
```
tkilias, line 99: Did you register this as pytest custom assert? Alternative, I would suggest, either rewrite it as MatcherObject or register it as custom assert.

tkilias: https://lorepirri.com/pytest-register-assert-rewrite.html
https://docs.pytest.org/en/7.1.x/reference/reference.html?highlight=pytest%20register_assert_rewrite#pytest.register_assert_rewrite
https://docs.pytest.org/en/7.1.x/how-to/assert.html?highlight=pytest%20register_assert_rewrite#assertion-introspection-details

exasol/analytics/audit/audit.py
ckunki: This is probably a misunderstanding. 
The create-statement for the Audit Table itself is never meant to be logged to the Audit Table.

The implementation of the `AuditQueryHandler` therefore does not (and should never) pass this query to `AudiTable.augment()`. Additionally, attribute `audit` to the constructor is set to the default value `False`.

exasol/analytics/audit/audit_query_handler.py
ckunki: see next push - I applied this change manually to avoid merge conflicts.

exasol/analytics/audit/audit_query_handler.py
ckunki: See next push.

exasol/analytics/audit/audit_query_handler.py
ckunki: Ah ok - see next push.

exasol/analytics/schema/table.py
ckunki: Ah - thanks. Great. 
See next push.

tests/unit_tests/audit_log/test_audit_query_handler.py
ckunki: `Mock()` requires the least amount of code, here. 
Also, the constructor will call `context.get_child_query_handler_context()`. 
So, an internal function or lambda is not sufficient from my perspective. 

Please correct me if I'm wrong.

ckunki: See next push.

tests/unit_tests/audit_log/test_with_python_runner.py
```diff
@@ -0,0 +1,109 @@
...
+    def start(self) -> Union[Continue, Finish[SampleOutput]]:
+        return Finish[SampleOutput](SampleOutput(self._parameter))
+
+    def handle_query_result(
+        self, query_result: QueryResult
+    ) -> Union[Continue, Finish[SampleOutput]]:
+        pass
+
+
+@contextlib.contextmanager
```
tkilias, line 64: why do you need a contextmanager, you are creating a function that doesn't use any resources

tests/unit_tests/query_handler_runner/test_python_query_handler_runner.py
```diff
@@ -393,12 +353,7 @@ def test_continue_query_list(aaf_pytest_db_schema, prefix, context_mock):
             """,
             MockResultSet(
                 rows=[(1,)],
-                columns=[
-                    Column(
-                        ColumnName("a"),
-                        ColumnType(name="DECIMAL", precision=1, scale=0),
-                    )
-                ],
+                columns=[decimal_column("a", precision=1, scale=0)],
```
tkilias, line 356: There was the urge too large simplify this, I guess

exasol/analytics/sql_executor/testing/mock_sql_executor.py
```diff
@@ -62,3 +62,19 @@ def execute(self, actual_query: str) -> ResultSet:
                 return next_expected_query.mock_result_set
             except StopIteration as e:
                 raise RuntimeError(f"No result set found for query {actual_query}")
+
+
+def expect_query(template: str, result_set=MockResultSet()):
```
tkilias, line 67: nice, this way we can it also use in other projects

exasol/analytics/audit/audit.py
```diff
@@ -42,14 +42,20 @@ def base_column_values(
     return {c.name.name: v for c, v in columns.items()}
 
 
+def uid_tester():
```
tkilias, line 45: I guess this is a leftover

tests/unit_tests/audit_log/test_with_python_runner.py
```diff
@@ -20,51 +18,59 @@
...
+    def start(self) -> Continue | Finish:
+        query = create_insert_query(TableNameImpl("T", SchemaName("S2")), audit=True)
+        return continue_action([query])
 
-    def handle_query_result(
-        self, query_result: QueryResult
-    ) -> Union[Continue, Finish[SampleOutput]]:
-        pass
+    def handle_query_result(self, result: QueryResult) -> Continue | Finish:
+        event_attributes = '{"c": 456}'
```
tkilias, line 66: maybe, we should have another round with a continue and an audit query here

tests/unit_tests/audit_log/test_with_python_runner.py
```diff
@@ -77,33 +83,160 @@ def factory(
...
+            FROM {view_fq};
+            """,
+            MockResultSet(rows=[(1,)], columns=[column]),
+        ),
+        expect_query(f"DROP VIEW IF EXISTS {view_fq};"),
+    ]
+
+
+@patch("exasol.analytics.audit.audit.uuid.uuid4")
+@patch("exasol.analytics.query_handler.query.select.uuid.uuid4")
```
tkilias, line 177: where does this come from? Is this from the log span?

tests/unit_tests/audit_log/test_with_python_runner.py
```diff
@@ -77,33 +83,160 @@ def factory(
...
-            top_level_query_handler_context=context_mock,
-            parameter=sample,
-            query_handler_factory=factory,
-        )
-        output = runner.run()
-    assert output.result == sample
+        prefix,
+        expect_query(""),
+        expect_query(""),
+    )
```
tkilias, line 186: I guess this is a leftover

tests/unit_tests/audit_log/test_with_python_runner.py
```diff
@@ -77,33 +83,160 @@ def factory(
...
+    audit_table = AuditTable(aaf_pytest_db_schema, prefix, run_id=Mock())
```
tkilias, line 187: you are mixing the prefix for tmp db objects with the table_prefix for the audit log, that is usually not a good idea. From the usage, you can probably use an arbitrary string for the table_prefix,

tests/unit_tests/audit_log/test_with_python_runner.py
```diff
@@ -127,16 +134,16 @@ def count_rows(event_name: str) -> list[str, MockResultSet]:
...
+              '{other_table.name}',
+              '{other_table.schema_name.name}',
               'TABLE',
               '{{{{"a": 123, "b": "value"}}}}',
               '{event_name}',
-              'UUID-2-2',
+              '{log_span_id}',
               'INSERT',
-              'UUID-2-1'\u0020
+              'RUN_ID_2'\u0020
```
tkilias, line 146: How does it come that we need \u0020 at the end

tkilias (approved):

tests/unit_tests/audit_log/test_with_python_runner.py
ckunki: Oh.
Long story.
It's rooted in method
[_insert_statement()](https://github.com/exasol/advanced-analytics-framework/blob/dccd31f905b7348a8ff3e7b1290488b331c2c24a/exasol/analytics/audit/audit.py#L220-L225) of `AuditTable`, currenty, implemented like this
```python
query_string=(
    f"INSERT INTO {self.name.fully_qualified} (\n"
    f"  {insert_statement.columns}\n"
    ") SELECT\n"
    f"  {insert_statement.values} {suffix}"
),
```

Previously, the suffix had been following directly the `insert_statement.values`, without space:
```python
    f"  {insert_statement.values}{suffix}"
```

The previous implementation requires the suffix (if not empty) to start with whitespace to separate its SQL from the `insert_statement.values`, which @ahsimb filed a review finding for.
So I agreed to insert the space in the f-string. But this, of course, impacts the tests with empty suffix to always add a space character at the end of each expected statement. Which IMHO is ugly and error-prone from a tests perspective.
And as `cleandoc()` will remove whitespace at eol (but also for readers and to avoid editors removing whitespace at eol, the space even needs to be specified as a unicode.

Please consider, that `_insert_statement()` is only an internal method in `AuditTable`.

In summary I see the following alternatives with the resp. disadvantages on production and testing code:

| ID | Alternative | Production code | Testing code and expectations |
|----|-------------|--------------|--------------|
| A1 | leave as it is | &#x2705; | ugly, error-prone |
| A2 | revert to the previous state | error if suffix is not empty, but does not start with whitespace | &#x2705; |
| A3 | prepend a space in `_insert_statement()` before suffix, only if suffix is not empty | (additional code) | &#x2705; |

ckunki (commented):
See comments

tkilias (approved):

tkilias (approved):

PR #285, opened 2025-04-17
Refactoring/#240 enhance column type

### LINKED ISSUE #217: Additional refactorings
xxx

### LINKED ISSUE #240: Improve the code quality of the schema.column_type.ColumnType
1. No documentation at all. Does the `name` include the size and precision? God knows.
2. If the `name` is not the full name, then the `full_name` property would be very helpful.
3. Consider implementing the `__eq__` method for type comparison.

Basically the attached PR replaces the isolated function calls `decimal_column()`, `varchar_column()`, etc. by classmethod `simple()` of classes `DecimalColumn`, `VarCharColumn`, etc.

The PR additionally introduces **dedicated classes** for each of the SQL column types in contrast to the approach before having only a generic class `Column` with attributes `name: ColumnName` and `type: ColumnType`.

Class `ColumnType` had attributes for all potential SQL column types without distinguishing which attribute was legal or even required for a specific SQL column type. Hence, it was possible to instantiate `ColumnType` with redundant, insufficient or even inconsistent attribute values.

Besides providing class method `simple()` for conveniently creating a column specifying its name as a simple `str` rather than an instance of `ColumnName` the PR **validates the arguments** wrt. the specific column class, e.g.
* Mandatory `size` for `VARCHAR`
* Number of Bits in the size of a `HASHTYPE` to be a multiple of 8
* Timestamp `WITH LOCAL TIME ZONE`
* Char set `ASCII` or `UTF8` for `CHAR` and `VARCHAR`
* Value ranges for `precision`, `scale`, `size`, etc. depending on the column type

Additionally, the PR unifies the **conversion** of columns between various formats
* Render a column for building a `CREATE TABLE` statement
* Parse an SQL definition like `VARCHAR(100) UTF8` and create an instance of a subclass of `Column`
* Create a Column instance based on the metadata returned by pyexasol
* Create a Column instance based on the metadata retrieved from `exa.meta` in a UTF

ckunki:
Closes #240


exasol/analytics/query_handler/query/result/udf_query_result.py
```diff
@@ -92,7 +90,8 @@ def _compute_columns(self, exa) -> List[Column]:
             column.name: column.sql_type for column in exa.meta.input_columns
         }
         columns = [
-            Column(ColumnName(key), ColumnType(column_dict[value]))
+            Column.from_sql_type(key, column_dict[value])
```
ckunki, line 93: Actually, this is location is unclear to me. 

1) The former implementation instantiated `ColumnType` with argument name set to `VARCHAR(20)`, for example.
AFAIK, column type names should not contain numeric arguments, as type Column also contained an attribute `size`, for example.

2) Additionally, these arguments are also not passed to `exa.meta.input_columns` in UDFs. 

3) Some tests used column types like `INTEGER` and `FLOAT`, which actually are mapped to `DECIMAL` and `DOUBLE` and hence cannot be reproduced in reality. 

All in all the former implementation and the related tests do not seem 100% plausible to me.

exasol/analytics/schema/column.py
```diff
@@ -1,76 +1,461 @@
-import dataclasses
-from typing import Optional
+# do not from __future__ import annotations
```
ckunki, line 1: Should I maybe split the large file column.py into
* `column/__init__.py`
* `column/boolean.py`
* `column/char.py`
* `column/date.py`
* `column/decimal.py`
* `column/double.py`
* `column/geometry.py`
* `column/hashtype.py`
* `column/timestamp.py`
* `column/varchar.py`

exasol/analytics/schema/column.py
ckunki, line 1: Should I maybe split the large file `column.py` into?
* `column/__init__.py`
* `column/boolean.py`
* `column/char.py`
* `column/date.py`
* `column/decimal.py`
* `column/double.py`
* `column/geometry.py`
* `column/hashtype.py`
* `column/timestamp.py`
* `column/varchar.py`

exasol/analytics/query_handler/query/result/udf_query_result.py
tkilias: I think, that happened out of lack of time, your implementation looks more correct

ckunki: OK, then let's go on with the current implementation and address potential issues as soon as we encounter these.

exasol/analytics/schema/column.py
```diff
@@ -1,76 +1,456 @@
...
+
+    def __post_init__(self):
+        """
+        As convenience Column supports specifying name as a simple str,
+        and __post_init__() instantiates ColumnName.
+        """
+        # If this approach is accepted, then we can remove classmethod
+        # simple() from the subclasses of Column which simplifies
+        # implementation and reduces overhead.
+        if isinstance(self.name, str):
```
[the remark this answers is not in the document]
ckunki: Damn - I verified and must admit you're right. 
Then for now, I propose to stay with method `.simple()`. :slightly_frowning_face:

doc/changes/unreleased.md
```diff
@@ -3,3 +3,7 @@
 ## Documentation
 
 * #283: Updated description and README
+
+## Refactorings
```
ckunki, line 7: Basically this PR replaces the isolated function calls `decimal_column()`, `varchar_column()`, etc. by class method `simple()` of classes `DecimalColumn`, `VarCharColumn`, etc.

The PR additionally introduces **dedicated classes** for each of the SQL column types in contrast to the approach before having only a generic class `Column` with attributes `name: ColumnName` and `type: ColumnType`.

Class `ColumnType` had attributes for all potential SQL column types without distinguishing which attribute was legal or even required for a specific SQL column type. Hence, it was possible to instantiate `ColumnType` with redundant, insufficient or even inconsistent attribute values.

Besides providing class method `simple()` for conveniently creating a column specifying its name as a simple `str` rather than an instance of `ColumnName` the PR **validates the arguments** wrt. the specific column class, e.g.
* Mandatory `size` for `VARCHAR`
* Number of Bits in the size of a `HASHTYPE` to be a multiple of 8
* Timestamp `WITH LOCAL TIME ZONE`
* Char set `ASCII` or `UTF8` for `CHAR` and `VARCHAR`
* Value ranges for `precision`, `scale`, `size`, etc. depending on the column type

Additionally, the PR unifies the **conversion** of columns between various formats
* Render a column for building a `CREATE TABLE` statement
* Parse an SQL definition like `VARCHAR(100) UTF8` and create an instance of a subclass of `Column`
* Create a Column instance based on the metadata returned by pyexasol
* Create a Column instance based on the metadata retrieved from `exa.meta` in a UTF

exasol/analytics/schema/column_types.py
```diff
@@ -0,0 +1,119 @@
...
+    Optional,
+)
+
+
+class CharSet(Enum):
+    UTF8 = auto()
+    ASCII = auto()
+
+    @classmethod
+    def from_string(cls, name: str) -> "CharSet":
```
ahsimb, line 19: Is `CharSet(name)` not good?

ahsimb, line 19: Is `CharSet(name)` not good?

exasol/analytics/schema/column_types.py
```diff
@@ -0,0 +1,119 @@
...
+    PRECISION = "precision"
+    SCALE = "scale"
+    SIZE = "size"
+    SRID = "srid"
+    UNIT = "unit"
+    WITH_LOCAL_TIME_ZONE = "withLocalTimeZone"
+
+
+@dataclass(frozen=True)
+class PyexasolMapping:
```
ahsimb, line 37: A docstring would help.

ahsimb: Why is it a dataclass?

exasol/analytics/schema/column_types.py
```diff
@@ -0,0 +1,119 @@
...
+    int_args: list[int]
+    modifier: str
+    source: ColumnTypeSource = ColumnTypeSource.SQL
+
+    def int_dict(self, keys: list[str]) -> dict[str, Any]:
+        return dict(zip(keys, self.int_args))
+
+    @property
+    def char_type_args(self) -> dict[str, Any]:
+        args: dict[str, Any] = dict(zip(["size"], self.int_args))
```
ahsimb, line 76: This is equivalent of
`args = {"size": self.int_args[0]} if self.int_args else {}`
I guess the case `not self.int_args` is invalid, so we should either raise an exception or just use `{"size": self.int_args[0]}`

exasol/analytics/schema/column_types.py
```diff
@@ -0,0 +1,119 @@
...
+        Instantiate SqlType based on its string representation, examples:
+
+        * "VARCHAR(100)"
+        * "DECIMAL(10,2)"
+        * "HASHTYPE(8 BIT)"
+        * "TIMESTAMP WITH LOCAL TIME ZONE"
+        * "CHAR ASCII"
+        """
+
+        raw = spec.strip().upper()
```
ahsimb, line 93: I would use re from the start to find all parts, especially since you are using it anyway.

exasol/analytics/schema/column_types.py
```diff
@@ -0,0 +1,119 @@
...
+
+        if name == "HASHTYPE" and " " in paren:
+            paren, modifier = paren.split()
+
+        int_args = [int(i) for i in paren.split(",")] if paren else []
+        return SqlType(name, int_args, modifier, ColumnTypeSource.SQL)
+
+    @classmethod
+    def from_pyexasol(cls, args: dict[str, Any], mapping: PyexasolMapping) -> "SqlType":
+        sql_name = args["type"]
```
ahsimb, line 115: Consider defining a string constant

exasol/analytics/schema/column_types.py
ckunki: Yes, you are right.
Thanks!
See next push, fixing this.

exasol/analytics/schema/column.py
```diff
@@ -1,76 +1,447 @@
...
+    PyexasolOption,
+    SqlType,
 )
 from exasol.analytics.utils.data_classes_runtime_type_check import check_dataclass_types
 
 
-@dataclasses.dataclass(frozen=True, repr=True, eq=True)
-class Column:
+class classproperty(property):
+    def __get__(self, owner_self, owner_cls):
```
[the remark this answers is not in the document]
ckunki: See next push.

exasol/analytics/schema/column.py
```diff
@@ -1,76 +1,447 @@
...
+        return self.fget(owner_cls)
+
+
+class ColumnClass(ABCMeta):
+    @abstractmethod
+    def sql_names(self) -> list[str]: ...
+
+    @abstractmethod
+    def from_sql(cls, column_name: str, sql_type: SqlType) -> "Column": ...
```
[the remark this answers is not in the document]
ckunki: See next push.

exasol/analytics/schema/column.py
```diff
@@ -1,76 +1,447 @@
...
+
+    def pyexasol_mapping(self) -> PyexasolMapping:
+        """
+        Default PyexasolMapping inherited as @classmethod by all classes
```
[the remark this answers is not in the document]
ckunki: no.

exasol/analytics/schema/column_types.py
ckunki: See next push providing a docstring.

ckunki: > Why is it a dataclass?

Maybe I don't understand the question? 

It's a dataclass to hold the attributes as defined. 
Or to put it differently: If you think it should _not_  be a dataclass, then why?

exasol/analytics/schema/column_types.py
ckunki: See next push.

exasol/analytics/schema/column_types.py
ckunki: I used re before and found it much shorter and easier to understand this way it is implemented currently.

exasol/analytics/schema/column_types.py
ckunki: See next push.

exasol/analytics/schema/column.py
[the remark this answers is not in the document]
ahsimb: So, what is inherited? `PyexasolMapping`?

exasol/analytics/schema/column.py
```diff
@@ -1,76 +1,447 @@
...
+    @property
+    def rendered(self) -> str:
+        return self.sql_spec(for_create=False)
+
+    @abstractmethod
+    def sql_spec(self, for_create: bool) -> str: ...
+
+    @classproperty
+    def sql_name(self):
+        return self.sql_names()[0]
```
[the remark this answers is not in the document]
ckunki: Thanks for enlightening discussion . 
The new solution is much simpler, does not need meta classes.
See next push.

exasol/analytics/schema/column.py
```diff
@@ -1,76 +1,447 @@
...
 
 
-@dataclasses.dataclass(frozen=True, repr=True, eq=True)
-class Column:
+class classproperty(property):
+    def __get__(self, owner_self, owner_cls):
+        return self.fget(owner_cls)
+
+
+class ColumnClass(ABCMeta):
```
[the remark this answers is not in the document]
ckunki: Thanks for enlightening discussion . 
The new solution is much simpler, does not need meta classes.
See next push.

exasol/analytics/schema/column.py
```diff
@@ -1,76 +1,447 @@
...
+
+    def sql_spec(self, for_create: bool = False) -> str:
+        return f"{self.sql_name}({self.size}) CHARACTER SET {self.charset.name}"
+
+    @classmethod
+    def sql_names(cls) -> list[str]:
+        return ["CHAR"]
+
+    @classmethod
+    def simple(
```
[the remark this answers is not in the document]
ckunki: It is called `simple()` because normally you need to provide an instance of `ColumnName` to the constructor, while the `simple()` method accepts a simple string:
```python 
VarCharColumn(ColumnName("A"), size=20)
```
vs.
```python 
VarCharColumn.simple("A", size=20)
```

I would have preferred to have a generic constructor accepting both data types, but this interferes with `@dataclass(frozen=True)` and I failed not fix this in a way accepted by mypy.

exasol/analytics/schema/column_types.py
ckunki: This comment helps. Thanks. 
See next push not using a dataclass anymore.

exasol/analytics/schema/column.py
```diff
@@ -1,76 +1,469 @@
...
+    SqlType,
 )
 from exasol.analytics.utils.data_classes_runtime_type_check import check_dataclass_types
 
 
-@dataclasses.dataclass(frozen=True, repr=True, eq=True)
-class Column:
+class UnsupportedSqlType(RuntimeError):
+    """
+    When calling ColumnClass.from_sql_name() with an name of an SQL column
```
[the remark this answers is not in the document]
ckunki: See next push.

exasol/analytics/schema/column.py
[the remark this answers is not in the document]
ckunki: Thanks for enlightening discussion.
See next push making `pyexasol_mapping()` an abstract classmethod of class `Column`.

exasol/analytics/schema/column.py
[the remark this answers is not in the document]
ahsimb: I see. As I commented in a different place, I suspect that the dataclass is a bit overused. Yes, every class can be declared as dataclass, but it doesn't mean it has to be. This one seems more like a functional class to me. The constructor is a functional constructor, not just a way to set up all fields.

exasol/analytics/schema/column.py
```diff
@@ -1,76 +1,447 @@
...
+
+    @abstractmethod
+    def sql_spec(self, for_create: bool) -> str: ...
+
+    @classproperty
+    def sql_name(self):
+        return self.sql_names()[0]
+
+    @classmethod
+    def check_arg(cls, name: str, value: int, allowed: range):
```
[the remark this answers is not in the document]
ckunki: I made it a class method to be able to include the name of the class in the exception.

exasol/analytics/schema/column.py
```diff
@@ -1,76 +1,447 @@
...
+
+    def pyexasol_mapping(self) -> PyexasolMapping:
+        """
+        Default PyexasolMapping inherited as @classmethod by all classes
+        declaring ColumnClass as metaclass, overriden in some classes.
+        """
+        return PyexasolMapping(int_keys=[], modifier_key=None)
+
+    @classmethod
+    def from_sql_name(cls, sql_name: str) -> "ColumnClass":
```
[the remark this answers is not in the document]
ckunki: I think you are right. I renamed method `from_sql_name()` to `get_class()`.

exasol/analytics/schema/column.py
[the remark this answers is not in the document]
ahsimb: Going back to the name, "simple" is not really self-explanatory. I can't think of a better name from the top of my head. The solution, as always, a docstring.

exasol/analytics/schema/column.py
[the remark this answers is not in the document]
ahsimb: The docstring is the same. I find it confusing.

tests/unit_tests/data_science_utils/schema/test_column.py
```diff
@@ -1,119 +1,264 @@
...
+    TimeStampColumn,
+    UnsupportedSqlType,
+    VarCharColumn,
+)
+from exasol.analytics.schema.column_types import (
+    PyexasolOption,
+    SqlType,
 )
 
+TEST_CASES = [
```
ahsimb, line 33: Can you define a dataclass for the test case? It will make the code easier to read.

exasol/analytics/schema/column.py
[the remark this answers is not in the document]
ckunki: See next push.

tests/unit_tests/data_science_utils/schema/test_column.py
ckunki: I don't understand the question.
The purpose of the variable TEST_CASES is to reuse the same list of test cases in multiple actual test cases:
* test_from_sql_spec
* test_for_create

exasol/analytics/schema/column.py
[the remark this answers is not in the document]
ckunki: Proposal: 
>        This method returns the default instance of PyexasolMapping. The
>        method pyexasol_mapping() is inherited as @classmethod by all
>        subclasses of Column, overriden in some of them.

ahsimb (approved):

PR #295, opened 2025-05-14
Refactoring/#294 refactor column name type

### LINKED ISSUE #292: Added check for range of size attribute to HashTypeColumn

### LINKED ISSUE #294: Refactor class Column once again

ckunki:
Closes #294 
Closes #292


exasol/analytics/schema/column.py
ckunki, line 1: Diff is quite large for this file.

ahsimb:
A general comment. The `column.py` is now mostly about the column type. Would you like to create a separate file `column_type.py` and move the type classes there? Also, I think naming should change. For example the `VarcharColumn` should be renamed to `VarcharColumnType` or `VarcharType`.

ahsimb:
I didn't notice first that the simple still creates a `Column`. It looks really weird when A contains B, but B can create A. I would strongly suggest that the column type minds its own business. So for example it can theoretically be used in something like `MyFancyColumn` which is not derived from `Column`.

exasol/analytics/schema/column.py
```diff
@@ -104,62 +91,72 @@ def get_class(cls, sql_name: str) -> type["Column"]:
...
+    than a ColumnName object.
+    """
+    from exasol.analytics.schema.column import Column
+
+    return Column(ColumnName(name), type)
 
 
 @dataclass(frozen=True, repr=True, eq=True)
-class BooleanColumn(Column):
+class BooleanColumn(ColumnType):
```
tkilias, line 132: if the super type is ColumType, I am not sure if BooleanColumn is an appropriate name , some for the other ones

exasol/analytics/schema/column.py
```diff
@@ -31,7 +27,7 @@ def __get__(self, owner_self, owner_cls):
 
 
 @dataclass(frozen=True, repr=True, eq=True)
-class Column:
+class ColumnType:
```
tkilias, line 30: would SQLType maybe more correct?

ckunki: Not too bad in my eyes. I only already have a class [SqlType in column_type_utils.py](https://github.com/exasol/advanced-analytics-framework/blob/b6bcb6e2045e02709a439e9c7d4b13be612d9105/exasol/analytics/schema/column_type_utils.py#L65) already, that should be renamed then.

PR #325, opened 2025-11-03
Refactoring/323 updated dependency to structlog

### LINKED ISSUE #323: Update structlog
Ticket  #321 updated the dependency declaration to pyexasol.
Updating other dependencies as well, made the CI build fail.
It turned out that `structlog` version 25.5 seems to have introduced a breaking change. 

Ticket #321 lists the tests that failed at that time. 
There are many other tests using structlog that did not fail.

The current ticket requests to
* [x] Update dependency structlog to version >= 25.5
* [x] Analyze and fix the problem, i.e. make AAF compatible with newer versions of structlog again

ckunki:
Closes #323

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1131,19 +1131,19 @@ vectorstore-mmr = ["numpy (>=1)", "simsimd (>=3)"]
 
 [[package]]
 name = "exasol-bucketfs"
-version = "1.1.0"
+version = "2.1.0"
 description = "BucketFS utilities for the Python programming language"
 optional = false
 python-versions = "<4.0,>=3.10"
 groups = ["main", "dev"]
 files = [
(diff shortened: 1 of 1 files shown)


poetry.lock
```diff
@@ -1131,19 +1131,19 @@ vectorstore-mmr = ["numpy (>=1)", "simsimd (>=3)"]
 
 [[package]]
 name = "exasol-bucketfs"
-version = "2.1.0"
+version = "1.1.0"
```
ArBridgeman: exasol-bucketfs went from 2.1.0 to 1.1.0
exasol-saas-api went from 2.4.0 to 1.1.1

Is this something that we'd want to happen?

It could indicate that some dependency is holding down a required transitive value. This could be problematic in the long-term if we allow up to version 3.0 in exasol-saas-api, but only test at 1.1.1

ArBridgeman (approved):

